PR #3, opened 2023-05-25
Added URL pattern matching

ofermend:
Updated website_crawler to include a URL pattern option (defaults to disabled), that defines a regular expression that needs to match the URL in order to index that content.

Updated documentation to reflect this


eskibars:
This is a fine first step.  There is a significant improvement that can be made though to drastically decrease crawl times of many websites if I'm understanding the code/structure correctly: https://github.com/vectara/vectara-ingest/blob/4eff217f387a43442638cd79afe1917aecc02d8c/crawlers/website_crawler.py#L35 doesn't take the regular expression into account, so it might navigate to many out-of-bounds pages that ultimately get thrown away by the regular expression.  Can fix that later

eskibars (approved):
LGTM.  Left 1 comment more generally on the PR

PR #5, opened 2023-05-30
Updates to docs crawler

ofermend:
- Changed docusaurus crawler to use playwright and crawling of docs instead of using the repo to avoid edge cases
- Improved index_url to support additional file types


config/langchain-docusaurus-search.yaml
```diff
@@ -0,0 +1,17 @@
+vectara:
```
ofermend, line 1: Added configuration for crawling all of langchain docs

config/vectara-docusaurus-search.yaml
```diff
@@ -7,5 +7,5 @@ crawling:
   crawler_type: docusaurus
 
 docusaurus_crawler:
-  docs_repo: "https://github.com/vectara/vectara-docs"
   docs_homepage: "https://docs.vectara.com/docs"
+  url_regex: [".*vectara.com/docs.*"]
```
ofermend, line 11: Parameters for docs crawl now support regex patterns

requirements.txt
```diff
@@ -1,4 +1,4 @@
-requests==2.28.2
+requests==2.31.0
```
ofermend, line 1: Upgrade to fix security vulnerability

justinhayes:
This all looks good to me. Although I don't have any experience with playwright, so for that part of it my review isn't worth all that much.

core/indexer.py
```diff
@@ -176,7 +217,9 @@ def index_url(self, url: str, metadata: dict) -> bool:
             soup = BeautifulSoup(content, 'html.parser')
             title = soup.title.string if soup.title else 'No Title'
             elements = partition_html(text=content)
-            elements = [str(e) for e in elements if type(e)==us.documents.elements.NarrativeText]
+
+        # only capture the text elements (as str) and ignore others
```
eskibars, line 221: Can you provide a bit more detail on why this needs to be done?

eskibars (approved):
LGTM.  Thanks!

core/indexer.py
ofermend: The idea is to only take narrative text elements and the str(e) converts them to an actual string.
See https://unstructured-io.github.io/unstructured/getting_started.html#document-elements for other element types. We can refine this later, but wanted to keep it simple to start.

PR #6, opened 2023-06-02
Added Retry and graceful fail for indexing.

ofermend:
Also added using requests.session so we don't create sessions on every call.


tallatshafaat (approved):
Thanks Ofer.
Regarding switching to session, one of our customers had a similar issue, and I guided them here: https://discuss.vectara.com/t/how-to-speed-queries-up/103

PR #8, opened 2023-06-21
improve crawling speed and new crawlers

ofermend:
- New crawler Arxiv: get top research papers by citations
- Including instructions to run on Windows and generally updated and simplified the getting started
- Indexer much improved: shorter time to retrieve pages using Playwright without pulling images to shorten load time, reusing playwright browser to reduce resource load, and more.
- After a lot of testing, using trafilatura to extract HTML content - it performs the best and removes the issue we had with hidden content.
- Various other general improvements and bug fixes.


README.md
```diff
@@ -30,13 +30,13 @@ Let’s create a basic crawler to scrape content from [Paul Graham's website](ht
...
 Install [python >= 3.8](https://www.python.org/downloads/) if it's not already installed.
 
-Install [yq](https://github.com/mikefarah/yq):
-
-- Mac: `brew install yq`
-- Linux: `snap install yq`
-
 Install [Docker](https://docs.docker.com/engine/install/).
 
+For Windows,
```
ofermend, line 35: Windows is complicated. The best way I found to install on windows is to use WSL and thus these instructions.

config/vectara-docs.yaml
```diff
@@ -7,5 +7,6 @@ crawling:
   crawler_type: docusaurus
 
 docusaurus_crawler:
-  docs_homepage: "https://docs.vectara.com/docs"
+  base_urls: ["https://docs.vectara.com/docs"]
   url_regex: [".*vectara.com/docs.*"]
+  extensions_to_ignore: ["php", "java", "py", "js"]
```
ofermend, line 12: Added ability to specify extensions to ignore in docs crawler.

core/indexer.py
```diff
@@ -82,10 +43,15 @@ def __init__(self, cfg: OmegaConf, endpoint: str, customer_id: str, corpus_id: i
...
         self.reindex = reindex
+        self.timeout = 30
 
         self.session = requests.Session()
-        adapter = requests.adapters.HTTPAdapter(max_retries=5)
+        adapter = requests.adapters.HTTPAdapter(max_retries=3)
         self.session.mount('http://', adapter)
+        self.session.mount('https://', adapter)
+
+        p = sync_playwright().start()
```
ofermend, line 53: This generates a browser instance that does not need to be recreated every time, reducing time to launch the browser every time, and saves memory resources too.

core/indexer.py
```diff
@@ -96,7 +62,32 @@ def _check_response(self, response) -> bool:
             return False
         else:
             return True
-        
+
+    def fetch_content_with_timeout(self, url: str, timeout: int = 30):
```
ofermend, line 66: Moved back to sync fetch since it's simpler and has the timeout variable.

core/indexer.py
```diff
@@ -96,7 +62,32 @@ def _check_response(self, response) -> bool:
...
+        '''
+        Fetch content from a URL with a timeout.
+        Args:
+            url (str): URL to fetch.
+            timeout (int, optional): Timeout in seconds. Defaults to 30.
+        Returns:
+            str: Content from the URL.
+        '''
+        page = self.browser.new_page()
+        page.route("**/*", lambda route: route.abort()  # do not load images as they are unnecessary for our purpose
```
ofermend, line 76: This results in images being ignored in the page, so that we don't spend time loading them. Results in 2-3x speedup.

crawlers/arxiv_crawler.py
```diff
@@ -0,0 +1,77 @@
+import requests
+import logging
+from core.crawler import Crawler
+import arxiv
+
+class ArxivCrawler(Crawler):
+
+    def get_citations(self, arxiv_id):
+        base_url = "https://api.semanticscholar.org/v1/paper/arXiv:"
```
ofermend, line 9: Arxiv does not have citations. so using semantic scholar to get that.

crawlers/discourse_crawler.py
```diff
@@ -33,12 +33,13 @@ def __init__(self, cfg: OmegaConf, endpoint: str, customer_id: str, corpus_id: i
         super().__init__(cfg, endpoint, customer_id, corpus_id, api_key)
         self.discourse_base_url = self.cfg.discourse_crawler.base_url
         self.discourse_api_key = self.cfg.discourse_crawler.discourse_api_key
+        self.session = requests.Session()
```
ofermend, line 36: Using requests.Session() so a session is not created every time.

crawlers/fmp_crawler.py
```diff
@@ -57,17 +54,21 @@ def crawl(self):
...
+                    response = self.session.get(url)
                 except Exception as e:
                     logging.info(f"Error getting transcript for {ticker} quarter {quarter} of {year}: {e}")
                     continue
                 if response.status_code == 200:
                     data = response.json()
                     doc_title = f"10-K for {company_name} from {year}"
-                    metadata = {'source': 'finsearch', 'title': doc_title, 'ticker': ticker, 'compnay name': company_name, 'year': year, 'type': '10-K'}
+                    rel_filings = [f for f in filings if f['acceptedDate'][:4] == str(year)]
+                    url = rel_filings[0]['finalLink'] if len(rel_filings)>0 else None
```
ofermend, line 70: bugfix: finalLink is the correct URL of the 10-K

crawlers/fmp_crawler.py
```diff
@@ -84,60 +85,13 @@ def crawl(self):
                                     document['section'].append({'title': f'{key} - {title}', 'text': '\n'.join(values)})
                     self.index_doc(document)
 
-
-            # income statements
```
ofermend, line 88: I didnt find these income_statement and cashflow_statement "Artificial" documents to help with anything, and they don't have a URL to point to as reference, so removing for now.

crawlers/website_crawler.py
```diff
@@ -21,8 +20,8 @@ def crawl(self):
         crawled_urls = set()
 
         if 'url_regex' in self.cfg.website_crawler:
+            url_regex = [re.compile(r) for r in self.cfg.website_crawler.url_regex]
```
ofermend, line 23: url_regex is now a list allowing multiple patterns

README.md
```diff
@@ -65,9 +65,11 @@ Make sure the Docker app is running.
 
 Then execute the run script from the root directory using your config file to tell it what to crawl and where to ingest the data, assigning the “default” profile from your secrets file:
 
+Start the crawler job:
 ```sh
 bash run.sh config/pg-rss.yaml default
 ```
+For Microsoft Windows, make sure you run this command from within the WSL environment (after running `wsl` to enter the linux environment).
```
cjcenizal, line 72: Nit: "Linux" should be capitalized

config/legalaid-IL.yaml
```diff
@@ -1,14 +1,14 @@
 vectara:
-  corpus_id: 141
+  corpus_id: 166
```
cjcenizal, line 2: Nit: I suggest lower-casing the name "legalaid-il" for consistency. Perhaps even writing it out would make it more readable: "legalaid-illinois"

core/indexer.py
cjcenizal: Would it be helpful to folks if this was a comment in the code?

crawlers/arxiv_crawler.py
cjcenizal: I think this would be a great code comment, because of the context it provides.

cjcenizal (approved):
I didn't test this stuff locally and I'm not very familiar with the codebase, so my review is fairly superficial. But from what I could understand these look like good improvements. Had a few minor suggestions.

crawlers/CRAWLERS.md
```diff
@@ -9,7 +9,19 @@
 `vectara-ingest` includes a number of crawlers that make it easy to crawl data sources and index the results into Vectara.
 Let's go through each of these crawlers to explain how they work and how to customize them to your needs. This will also provide good background to creating (and contributing) new types of crawlers.
 
-### website crawler
+## Supported Integrations
+<table>
+  <tr>
```
eskibars, line 14: Can you please just double check that all of these icons are licensed for public use

crawlers/arxiv_crawler.py
```diff
@@ -0,0 +1,117 @@
...
+                    'title': result.title,
+                    'authors': result.authors,
+                    'abstract': result.summary,
+                    'published': str(date)
+                })
+        except Exception as e:
+            logging.info(f"Exception {e}, we have {len(papers)} papers already, so will continue with indexing")
+
+        # sort by citation count and get top n papers
+        if self.cfg.arxiv_crawler.sort_by == 'citations':
```
eskibars, line 106: Should we throw an exception if, after filtering, papers is an empty array?

core/indexer.py
```diff
@@ -1,74 +1,36 @@
...
-from requests.adapters import HTTPAdapter, Retry
-
+import time
 import asyncio
-from unstructured.partition.html import partition_html
-from unstructured.partition.pdf import partition_pdf
-import unstructured as us
+from core.utils import create_session_with_retries
+
+from trafilatura import extract, extract_metadata
```
eskibars, line 9: trafilatura is licensed under GPL 3, and is thus not legally compatible with the Apache 2 license.  We must drop it

crawlers/docusaurus_crawler.py
```diff
@@ -1,66 +1,59 @@
...
 
     def crawl(self):
-        self.crawled = set()
-        self.ignored = set()
-        self.site_url = self.cfg.docusaurus_crawler.docs_homepage
-        self.domain_regex = [re.compile(r) for r in self.cfg.docusaurus_crawler.url_regex]
-        if self.site_url.endswith('/'):
-            self.site_url = self.site_url[:-1]
-        self.crawl_urls([self.site_url])
+        self.extensions_to_ignore = list(set(self.cfg.docusaurus_crawler.extensions_to_ignore + ["png", "svg", "jpeg", "jpg", "gif", "mp4"]))
```
eskibars, line 50: Can you sort these alphabetically?  Eventually, we should make these configurable

crawlers/pmc_crawler.py
```diff
@@ -4,11 +4,10 @@
 import json
 from lxml import etree
```
eskibars, line 5: I just noticed this.  We should fix this quickly by replacing this with a better XML parser: lxml/etree has serious security issues

requirements.txt
```diff
@@ -25,15 +25,20 @@ boto3==1.26.116
...
 nbconvert==7.2.10
 nbformat==5.8.0
-docutils==0.20.1
\ No newline at end of file
+docutils==0.20.1
+arxiv==1.4.7
+trafilatura==1.5.0
+sqlalchemy==2.0.14
+pymysql==1.0.3
+psycopg2-binary==2.9.6
```
eskibars, line 44: psycopg2-binary is LGPL licensed and we should remove it from the dependencies

eskibars (changes requested):

config/legalaid-IL.yaml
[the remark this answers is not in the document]
ofermend: Since no API keys are provided, it's not a real security risk. Am I wrong?
If we remove this it becomes difficult to include examples. I really use these all the time (e.g. for regression) and in this case would need to copy them in/out all the time.

config/vectara-jira.yaml
```diff
@@ -7,7 +7,6 @@ crawling:
   crawler_type: jira
 
 jira_crawler:
-  jira_base_url: "https://zir-ai.atlassian.net/"
+  jira_base_url: "https://vectara.atlassian.net/"
```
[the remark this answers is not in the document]
ofermend: Again, similar to corpus ID, if I do this, then I can't use it internally.
This is doable in general, since we now use the docker-hub and I can probably create that outside. But if working on an internal PR (docker-hub not yet updated) it becomes super annoying and error prone to test.

crawlers/pmc_crawler.py
ofermend: compliments I think of GPT4. Can you pls share what the issues are and what is recommneded to use instead?

PR #21, opened 2023-07-18
Kareema branch

KareemaBatool:


config/readthedocs.yaml
```diff
@@ -0,0 +1,14 @@
+vectara:
```
ofermend, line 1: I would rename this file to "azureml.yaml" and not "readthedocs.yaml" since it really crawls the azureml documentation. (readthedocs is generic)

config/vectara-docs.yaml
```diff
@@ -1,12 +1,13 @@
 vectara:
-  corpus_id: 140
-  customer_id: 1526022105
+  corpus_id: 2
```
ofermend, line 2: I realize you had to change this to your account and corpus_id, but please revert back to the original numbers before we merge with main.

config/vectara-notion-example.yaml
```diff
@@ -0,0 +1,8 @@
+vectara:
```
ofermend, line 1: How is this different than the "vectara-notion.yaml" that already exists? Remember this is a public github repo so think about potential confusion from users. I would recommend keeping the original since you really did not make any changes to the notion crawler or the file (except trying it out as an example).

config/vectara-docs.yaml
```diff
@@ -1,12 +1,13 @@
...
+  customer_id: 2081868821
   reindex: true
 
 crawling:
-  crawler_type: docusaurus
+  crawler_type: doc
+  document_type: docusaurus
```
ofermend, line 8: I think this fits better inside the "docs_crawler" section, since it is specific to the docs_crawler process.

crawlers/CRAWLERS.md
```diff
@@ -108,7 +108,7 @@ The RSS crawler can be used to crawl URLs listed in RSS feeds such as on news si
 - `delay` defines the number of seconds between to wait between news articles, so as to make the crawl more friendly to the hosting site.
 - `extraction` defines how we process URLs (pdf or html) as as with the website crawler.
 
-### Docusaurus crawler
+### Docs crawler
```
ofermend, line 111: I would rewrite this a bit in the following way.
- First, the yaml should say "docs_crawler" not "docusaurus_crawler".
- Similarly, "The Docusaurus crawler processes..." need to be "The Docs crawler processes..."
- You can explain in words that this crawler supports two types of documentation systems: readthedocs and docusaurus. The parameters are the same in each case - docs_repo and docs_homepage, and you can modify the "document_type" to differentiate between the two.

crawlers/doc_crawler.py
```diff
@@ -7,7 +7,7 @@
 
 from core.utils import create_session_with_retries
 
-class DocusaurusCrawler(Crawler):    
+class DocCrawler(Crawler):
```
ofermend, line 10: I thought we said we'll rename this to DocsCrawler (and if so, also rename the file to docs_crawler.py)

crawlers/doc_crawler.py
```diff
@@ -47,13 +47,23 @@ def get_urls(self, base_urls):
         return crawled_urls
 
     def crawl(self):
-        self.extensions_to_ignore = list(set(self.cfg.docusaurus_crawler.extensions_to_ignore + ["gif", "jpeg", "jpg", "mp3", "mp4", "png", "svg"]))
-        all_urls = self.get_urls(self.cfg.docusaurus_crawler.base_urls)
-
-        url_regex = [re.compile(r) for r in self.cfg.docusaurus_crawler.url_regex]
+        if self.cfg.crawling.document_type == "docusaurus":
```
ofermend, line 50: I don't understand the need for the if/else here - I think the code needs to be the same (extensions to ignore, all_urls, url_regex, etc) the only differnece is the source, which should actually be read from the configuration file (document_type), no?

crawlers/hubspot_crawler
```diff
@@ -0,0 +1,50 @@
+import logging
```
ofermend, line 1: Can we make the hubspot changes in a separate branch (so this file should not be here) please?
this way after everything looks good with the changes for docs_crawler we can just merge that, and don't have cross-dependencies with hubspot which may take longer.

ingest.py
```diff
@@ -91,6 +91,9 @@ def main():
     env_dict = env_dict[profile_name]
 
     for k,v in env_dict.items():
+        if k=='HUBSPOT_API_KEY':
```
ofermend, line 94: Same comment as above re: Hubspot being in a seaprate branch and ultimately separate PR.

ofermend (changes requested):
Hey Kareema - please see my suggested improvements and let me take another look after those have been implemented.

PR #24, opened 2023-07-24
Hubspot crawler

KareemaBatool:


crawlers/CRAWLERS.md
```diff
@@ -185,6 +185,10 @@ The JIRA crawler indexes issues and comments into Vectara.
 The Notion crawler has no specific parameters, except the `NOTION_API_KEY` that needs to be specified in the `secrets.toml` file. 
 The crawler will index any content that is associated with this key that is on your notion instance.
 
+### Hubspot crawler
+
+The HubSpot crawler has no specific parameters, except the `HUBSPOT_API_KEY` that needs to be specified in the `secrets.toml` file. The crawler will index any content that is associated with this key that is on your Hubspot instance. The crawler also uses `clean_email_text()` module which takes the email message as a parameter and cleans it to make it more presentable. This function in `utils.py` can be extended to handle other indentation or cleaning problems. Currently, It is taking care of indentation character `>`.
```
ofermend, line 190: A few thoughts:
- "The crawler will index any content that is associated with this key that is on your Hubspot instance" - maybe worth adding that it's only email content. (hubspot has other content we're not indexing at this point)
- The comment about "the function in utils.py can be extended..." - that whole thing is a bit out of context for someone reading the specifics of the hubspot crawler. Perhaps remove it from the docs and just add the documtation directly in the code for that function?

crawlers/hubspot_crawler.py
```diff
@@ -0,0 +1,113 @@
...
+        headers = {
+            "Authorization": f"Bearer {self.hubspot_api_key}",
+            "Content-Type": "application/json"
+        }
+
+        properties = ["hs_email_subject", "hs_email_text"]
+
+        all_emails = []
+
+        after = 1
```
ofermend, line 31: Can you confirm - is the API starting from 1 (in which case this is okay) or 0 (in which case we're missing one email here)

crawlers/hubspot_crawler.py
```diff
@@ -0,0 +1,113 @@
...
+            "Subject": email_subjects,
+            "Text": email_texts
+        })
+
+        logging.info(f"Retrieved {len(email_df)} emails from HubSpot.")
+        return all_emails
+
+    def crawl(self):
+        logging.info("Starting HubSpot Crawler.")
+        all_emails = self.retrieve_emails_from_hubspot()
```
ofermend, line 77: This is probably okay for the first version, but one concern I have here is that we first take out ALL emails and then index them. If the resulting dataframe is very very large (lots of emails) this may become a problem. We can alternatively continuously get the emails and index them at the same time, without the need to store temporarily ALL of them in memory.

crawlers/hubspot_crawler.py
```diff
@@ -0,0 +1,113 @@
+import logging
+from core.crawler import Crawler
+from omegaconf import OmegaConf
+import requests
+import pandas as pd
+import re
+from core.indexer import Indexer
+from core.utils import clean_email_text
```
ofermend, line 8: I don't see this function in core/utils as part of the PR. Where is it?

PR #29, opened 2023-08-23
Other languages integration

ofermend:


core/utils.py
```diff
@@ -53,3 +56,18 @@ def clean_email_text(text):
     cleaned_text = re.sub(r"[<>]+", "", cleaned_text, flags=re.MULTILINE)
     return cleaned_text
 
+def detect_language(text):
```
ofermend, line 59: @KareemaBatool - detect_language can be a generic function. Why are you limiting its output to "ar", "ko", "zh" and "en" only? That is a limitation (I think) of Goose that it only needs to know about those languages. So maybe do that logic in the code that deals with Goose?

crawlers/CRAWLERS.md
```diff
@@ -9,6 +9,10 @@
 `vectara-ingest` includes a number of crawlers that make it easy to crawl data sources and index the results into Vectara.
 Let's go through each of these crawlers to explain how they work and how to customize them to your needs. This will also provide good background to creating (and contributing) new types of crawlers.
 
+### Multilingual Web Content Crawling and Indexing
+
+We support crawling and indexing web content in 42 languages currently. To determine the language of a given webpage, we utilize https://pypi.org/project/langdetect/. To cater to a larger number of languages we use both https://pypi.org/project/goose3/ and https://pypi.org/project/jusText/ to get all the web content. 
+
```
ofermend, line 15: I would say the last piece a bit differently. The use of Goose3+justext is (in my mind) not part of catering to multiple languages. Remember, it was a case I saw where Goose3 returned an empty text (in an English page) and Justext returns something there. So it's more that using both gives us better coverage of the text without suffering through all the links/ads and extra material that is often on a webpage and not relevant.
So perhaps move that piece to another part of the documentation, and just mention that it's used in indexer.index_url for any crawler that uses that method.

Issue #15, opened 2023-07-12
GitHub crawler: add pull requests


eskibars:
Many times, there's very rich text content in GitHub PRs in addition to the issues.  These would be useful to capture in Vectara as well

PR #37, opened 2023-10-01
improvements to github crawler

ofermend:
added PRs to github crawler (per issue #15)
Refactor to simplify

### DIFF
--- a/config/lethain.yaml
+++ b/config/lethain.yaml
@@ -1,5 +1,5 @@
 vectara:
-  corpus_id: 160
+  corpus_id: 240
   customer_id: 1526022105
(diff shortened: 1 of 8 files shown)


justinhayes:
Approved. I'm not familiar with the github API or object model, but your usage of that, as well as the vectara API, look good to me. One thing to consider (and it's larger in scope than this specific PR) is to add timestamp fields that are numeric (e.g. number of epoch seconds) whenever you have created_at or updated_at metadata fields. That way the user can do date range based filtering by defining the fields as filter attributes. But that's a nice to have, and not a blocker for this PR.

PR #45, opened 2023-11-10
Updates to CSV/DB crawlers

ofermend:
Allow to have a title per row when processing CSV files or DB files.


cjcenizal (approved):
@ofermend Have you tested this locally? Code LGTM overall but I'm not very familiar with this codebase. I also didn't test this.

crawlers/database_crawler.py
```diff
@@ -1,20 +1,48 @@
...
+            query = f'SELECT {",".join(all_columns)} FROM {db_table}'
+
+        conn = sqlalchemy.create_engine(db_url).connect()
+        df = pd.read_sql_query(sqlalchemy.text(query), conn)
+        logging.info(f"indexing {len(df)} rows from the database using query: '{query}'")
+
+        doc_id_columns = list(self.cfg.database_crawler.get("doc_id_columns", None))
+        self.index_dataframe(df, text_columns, title_column, metadata_columns, doc_id_columns)
+
+    def crawl_old(self) -> None:
```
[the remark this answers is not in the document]
ofermend: yes good catch - removed.

crawlers/csv_crawler.py
```diff
@@ -5,28 +5,26 @@
...
 
-    def crawl(self) -> None:
-        text_columns = list(self.cfg.csv_crawler.text_columns)
-        metadata_columns = list(self.cfg.csv_crawler.metadata_columns)
-        csv_path = self.cfg.csv_crawler.csv_path
-        csv_file = '/home/vectara/data/file.csv'
-        doc_id_columns = list(self.cfg.csv_crawler.get("doc_id_columns", None))
-        sep = self.cfg.csv_crawler.get("separator", ",")
-
+    def index_dataframe(self, df: pd.DataFrame, text_columns, title_column, metadata_columns, doc_id_columns) -> None:
```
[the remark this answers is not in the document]
ofermend: It's from the Pandas python package. In a nutshell it's like a locally stored table of variables. 
From ChatGPT: "A Pandas DataFrame is a two-dimensional, size-mutable, and potentially heterogeneous tabular data structure with labeled axes (rows and columns) in Python"

crawlers/csv_crawler.py
```diff
@@ -19,14 +20,17 @@ def crawl(self) -> None:
         logging.info(f"indexing {len(df)} rows from the CSV file {csv_path}")
 
         def index_df(doc_id: str, title: str, df: pd.DataFrame) -> None:
-            parts = []
+            texts = []
+            titles = []
             metadatas = []
             for _, row in df.iterrows():
+                titles.append(row[title_column])
```
justinhayes: I didn't test this locally, but I wanted to confirm that row[title_column] will result in a correct outcome regardless of whether the user provides a title_column config or not. Based on my read of the code this expects titles, texts, and metadatas to all have the same length... so assuming that is correct then will titles have the right length if title_column was never provided?

justinhayes: Actually, I think you might have changed this code around in a later commit. Either way, just a reminder to double check the logic here.

justinhayes:
I left one comment (which might be a nop). Once that is addressed then I approve.

crawlers/csv_crawler.py
ofermend: yes it's been an issue and as you saw fixed!

PR #46, opened 2023-11-20
fix(vectara-ingest): added the code to get the api endpoint from the config

irfanazam1:
Currently, the api endpoint is hardcoded to api.vectara.io. Changed the code to use the endpoint form the config, if present, else use the hardcoded default value.


ofermend (approved):
LGTM

PR #55, opened 2023-12-13
Website crawler fix

ofermend:
Update to usage of playwright to address issues with crawling.

### DIFF
--- a/core/crawler.py
+++ b/core/crawler.py
@@ -17,35 +17,31 @@ get_headers = {
     "Connection": "keep-alive",
 }
 
-def recursive_crawl(url: str, depth: int, url_regex: List[Any], visited: Optional[Set[str]]=None, session: Optional[requests.Session]=None) -> Set[str]:
+            
+def recursive_crawl(url: str, depth: int, url_regex: List[Any], indexer: Indexer, visited: Optional[Set[str]]=None) -> Set[str]:
     if depth <= 0:
         return set() if visited is None else set(visited)
 
     if visited is None:
         visited = set()
-    if session is None:
-        session = requests.Session()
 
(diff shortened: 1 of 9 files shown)


core/indexer.py
```diff
@@ -274,7 +274,7 @@ def index_url(self, url: str, metadata: Dict[str, Any]) -> bool:
         # for everything else, use PlayWright as we may want it to render JS on the page before reading the content
         else:
             try:
-                actual_url, html_content = self.fetch_content_with_timeout(url)
+                actual_url, html_content = self.fetch_content_with_timeout(add_www(url))
```
justinhayes: This appears to always force a 'www' onto a URL. But it's normal for URLs to not have www (e.g. https://docs.vectara.com/). Is this going to break those URLs?

core/crawler.py
```diff
@@ -17,35 +17,31 @@
...
-        session = requests.Session()
 
     # For binary files - we don't extract links from them, nor are they included in the crawled URLs list
-    # for document files (like PPT, DOCX, etc) we don't extract links from the, but they ARE included in the crawled URLs list
     url_without_fragment = url.split("#")[0]
     if any([url_without_fragment.endswith(ext) for ext in binary_extensions]):
         return visited
     visited.add(url)
+
+    # for document files (like PPT, DOCX, etc) we don't extract links from the, but they ARE included in the crawled URLs list
```
justinhayes: Typo in the comment - "... don't extract links from **the,** but ..."

justinhayes:
Added a few comments. Other than that it looks good to me (although I am not very familiar with the playwright stuff).

core/indexer.py
ofermend: I think you were looking at an older branch somehow? This is no longer in the code.

PR #60, opened 2024-01-04
Table processing option

ofermend:
This PR adds an optional step for processing tables in PDF documents before ingestion.
This is a relatively common approach in LangChain (e.g. https://github.com/langchain-ai/langchain/blob/master/cookbook/Semi_Structured_RAG.ipynb) or LlamaIndex:
1. Extract table (using unstructured.io) contents
2. Ask OpenAI to summarize the contents of that table as text, and ingest that text into the corpus, in addition to the other content we already ingest.

With this PR this is an option that can be turned on, and if so required an addition of OpenAI key
Note that this process is relatively slow (at least on my Mac M2) and so may or may not be useful at very large scale.


core/indexer.py
```diff
@@ -50,6 +50,11 @@ def __init__(self, cfg: OmegaConf, endpoint: str, customer_id: str, corpus_id: i
         self.timeout = cfg.vectara.get("timeout", 60)
         self.detected_language: Optional[str] = None
 
+        self.summarize_tables = cfg.vectara.get("summarize_tables", False)
+        if cfg.vectara.get("openai_api_key", None) is None:
+            self.summarize_tables = False
+            logging.info("OpenAI API key not found, disabling table summarization")
```
justinhayes, line 56: I would only emit this log line if the OpenAI api key is not found AND ALSO summarize_tables is true. If a user is not wanting to do summarize tables, then they won't add an OpenAI key, but if they see this info message it might be confusing to them because they didn't want table summarization in the first place.

Tonymzaramo (commented):
# timeout (optional); sets the URL crawling timeout in seconds

PR #69, opened 2024-03-14
added slack crawler

adeelehsan:


crawlers/slack_crawler.py
```diff
@@ -0,0 +1,179 @@
...
+        for message in messages:
+            text = message.get("text", "")
+            # Replace user IDs with usernames if mentioned in the message
+            if "<@" in text:
+                for uid in users_info:
+                    if uid in text:
+                        username = users_info[uid]
+                        text = text.replace(f"<@{uid}>", f"@{username}")
+                        message["text"] = text
+    except SlackApiError as e:
```
ofermend, line 59: What could be issues this exception occurs? Does it make sense to put the try/except at the message level, and then if an exception occurs just ignore replacing for that message (as opposed to all messages)?

crawlers/slack_crawler.py
```diff
@@ -0,0 +1,179 @@
...
+class SlackCrawler(Crawler):
+    def __init__(self, cfg: OmegaConf, endpoint: str, customer_id: str, corpus_id: int, api_key: str):
+        super().__init__(cfg, endpoint, customer_id, corpus_id, api_key)
+        self.user_token = self.cfg.slack_crawler.slack_user_token
+        self.client = WebClient(token=self.user_token)
+        self.days_past = self.cfg.slack_crawler.days_past
+        self.channels_to_skip = self.cfg.slack_crawler.channels_to_skip
+
+    def get_users_info(self):
+        while True:
```
ofermend, line 72: Why do you need this "While" here?

crawlers/slack_crawler.py
```diff
@@ -0,0 +1,179 @@
...
+
+                return users_info
+            except IncompleteRead as e:
+                logging.error(f"Error: {e}")
+
+    def get_channels(self):
+        """Returns the list of the channels of the workspace."""
+
+        channels = []
+        while True:
```
ofermend, line 88: Same question here: why is the While needed?

crawlers/slack_crawler.py
```diff
@@ -0,0 +1,179 @@
...
+    except SlackApiError as e:
+        logging.error(f"Error replacing user id's with user handlers: {e}")
+
+
+class SlackCrawler(Crawler):
+    def __init__(self, cfg: OmegaConf, endpoint: str, customer_id: str, corpus_id: int, api_key: str):
+        super().__init__(cfg, endpoint, customer_id, corpus_id, api_key)
+        self.user_token = self.cfg.slack_crawler.slack_user_token
+        self.client = WebClient(token=self.user_token)
+        self.days_past = self.cfg.slack_crawler.days_past
```
ofermend, line 68: Maybe replace with "self.cfg.slack_crawler.get("days_past", None)
So that if it's undefined will go to None

crawlers/slack_crawler.py
```diff
@@ -0,0 +1,179 @@
...
+        messages = []
+        cursor = None
+        last_message_timestamp = None
+        if self.days_past is not None:
+            last_message_timestamp = str(get_timestamp(self.days_past))
+        while True:
+            try:
+                response = self.client.conversations_history(channel=channel_id, oldest=last_message_timestamp,
+                                                             cursor=cursor,
+                                                             limit=200)
```
ofermend, line 109: What is the implication of this value of limit=200?
Does this mean it reads 200 messages at a time? If so, what is the implication of increasing/reducing it?
Not sure if it's meaningful to play with, but at least let's put it as a parameter above (instead of just 200 here hidden in the code) and document what it means.

crawlers/slack_crawler.py
```diff
@@ -0,0 +1,179 @@
...
+                messages += response["messages"]
+                # Check if there are more messages to retrieve
+                if not response["has_more"]:
+                    break
+                cursor = response["response_metadata"]["next_cursor"]
+            except IncompleteRead as e:
+                logging.error(f"IncompleteRead error occurred: {e}")
```
ofermend, line 116: Can u pls explain what is your intent here - what should happen in an exception case?
are we stopping the read? ignoring and getting the rest of the messages? etc?

crawlers/slack_crawler.py
```diff
@@ -0,0 +1,179 @@
...
+
+        return replies
+
+    def crawl(self) -> None:
+        try:
+            users_info = self.get_users_info()
+            channels = self.get_channels()
+            for channel in channels:
+                channel_id = channel["id"]
+                if channel["name"] in self.channels_to_skip:
```
ofermend, line 142: Does it make sense to add a check at the start of the class to make sure all the channels listed in channels_to_skip actually exist in the Slack client - so that if someone intends to exclude a channel and mistypes its name, there is a warning message?

crawlers/slack_crawler.py
```diff
@@ -0,0 +1,179 @@
...
+                                continue
+
+                    if doc_text is None and msg.get("subtype") == "bot_message":
+                        for attachment in msg.get("attachments", []):
+                            doc_text = f'{attachment.get("text", "")}\n'
+
+                    document = {
+                        "documentId": doc_id,
+                        "metadataJson": doc_metadata,
+                        "description": doc_text,
```
ofermend, line 172: Wait so looks like you are using msg['text'] as a description of the text instead of the first section. Can you please elaborate why that is? Isn't the msg text (before replies start) an integral part of the text and needs to be in the section part?

crawlers/slack_crawler.py
adeelehsan: Limit represent the number of messages would be return in a api call. By default it is 100 and maximum is 999. Bu they have mentioned in the docs the 200 is optimal value. Having a higher number of message throws error sometime e.g IncompleteRequest or timeout.

crawlers/slack_crawler.py
ofermend: I would recommend dealing with this in a more robust way. Catch that exception and retry e.g up to X times and then throw an error. Right now, if this happens all the time the code will just hang and the user won't know what is going on...

crawlers/slack_crawler.py
ofermend: Cool. So suggesting you make it a variable and just document why (maybe link to the docs recommending 200)

ofermend:
Why is poetry.lock conflicting?
otherwise LGTM. Did not test locally.

ofermend (approved):

PR #81, opened 2024-04-14
Fixing issue of Ray with ARM...

ofermend:
updated CRAWLERS.MD with documentation on arguments for docs_Crawler
Updated Docker process to work better with ARM vs X86 architecture
Now the docker generated matches the architecture so Ray can work properly.


adeelehsan (approved):
Looks good. Did tested for the ARM and its working. 
Did't test for AMD

PR #82, opened 2024-04-15
Update to rate limiting functionality and a few more things

ofermend:
Removed rate_limiter package and did internal implementation to avoid using an old and unmaintained package
Updates to documentation
Fixed issue with logging with Ray
Updated rate limiting configuration


Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN apt-get update
 
 ENV HOME /home/vectara
 ENV XDG_RUNTIME_DIR=/tmp
+ENV RAY_DEDUP_LOGS="0"
```
ofermend, line 18: This ensures full logging with ray so it does not remove semi-dup msgs

README.md
```diff
@@ -179,7 +179,10 @@ vectara:
   reindex: false
   
   # timeout: sets the URL crawling timeout in seconds (optional)
-  timeout:60
+  timeout: 90
```
ofermend, line 182: Changed to fit the actual default in the code

config/askFeynman.yaml
```diff
@@ -8,7 +8,7 @@ crawling:
 
 website_crawler:
   urls: ["https://www.feynmanlectures.caltech.edu/"]
-  delay: 1
+  num_per_second: 1
```
ofermend, line 11: Changing all rate limiting configuration to be using "num_per_second" - specifying an integer, so that calls would not be more than X/second.

core/indexer.py
```diff
@@ -146,20 +148,22 @@ def fetch_page_contents(self, url: str, debug: bool = False) -> Tuple[str, str,
                 else route.continue_() 
             ) 
             if debug:
-                page.on('console', lambda msg: logging.info(f"playwright debug: {msg.text})"))
+                page.on('console', lambda msg: self.logger.info(f"playwright debug: {msg.text})"))
 
-            page.goto(url, timeout=self.timeout*1000, wait_until="networkidle")
+            page.goto(url, timeout=self.timeout*1000, wait_until="load")
```
ofermend, line 153: This is based on feedback from the hackathon - seems like networkidle is problematic in some cases where a webpage continues to update, resulting in a timeout. This should be more accurate.

adeelehsan (approved):

PR #84, opened 2024-04-21
remove attrdict

ofermend:
fixed regression due to python version upgrade


core/utils.py
```diff
@@ -86,21 +86,17 @@ def remove_anchor(url: str) -> str:
...
     """Normalize a URL by removing query parameters."""    
     # Prepend with 'http://' if URL has no scheme
     if '://' not in url:
         url = 'http://' + url
     p = urlparse(url)
-        
-    # Remove query parameters
-    path = p.path.split('?', 1)[0]
+    query = p.query if keep_query_params else ''
+    return ParseResult(p.scheme, p.netloc, p.path, '', query, '').geturl()
```
justinhayes, line 96: Does the `p.path` in this line need to instead be `p.path.split('?', 1)[0]` which is what was used in the previous version of this function? Or does ParseResult take care of splitting off just the path for you?

justinhayes (approved):

core/utils.py
ofermend: The previous code was actually wrong. the p.path.split('?', 1)[0] did not work as intended. The new code now works properly.

PR #86, opened 2024-05-05
Option to remove docs not crawled

ofermend:
Added remove_old_content configuration option.
If enabled, if will use the list-documents API to list all documents in the corpus crawled, and then compare that against the list of documents crawled in this session, while removing any previously indexed documents that are no longer there. THis helps when re-crawling of the entire corpus occurs periodically where some pages may be deprecated by the source (e.g. in website or docs crawling)

### DIFF
--- a/core/indexer.py
+++ b/core/indexer.py
@@ -192,6 +192,47 @@ class Indexer(object):
             return False
         return True
     
+    def _list_docs(self) -> List[Dict[str, str]]:
+        """
+        List documents in the corpus.
+        Returns:
+            list of [docId, URL] for each document in the corpus
+            URL taken from metadata if exists, otherwise it'd be None
+        """
+        page_key = None  # Initialize page_key as None
+        docs = []
+    
+        # Loop until there's no next page
+        while True:
+            body = {"corpusId": self.corpus_id, "numResults": 100}
+            if page_key:  # Add page_key to the request if it's not None
+                body["pageKey"] = page_key
+
+            post_headers = { 
+                'x-api-key': self.api_key, 
+                'customer-id': str(self.customer_id), 
+                'X-Source': self.x_source
+            }
+            response = self.session.post(
+                f"https://{self.endpoint}/v1/list-documents", data=json.dumps(body),
+                verify=True, headers=post_headers)
+            if response.status_code != 200:
+                self.logger.error(f"Error listing documents with status code {response.status_code}")
+                return []
+            res = response.json()
+
+            # Extract URLs from documents
+            for doc in res['document']:
+                url = next((md['value'] for md in doc['metadata'] if md['name'] == 'url'), None)
+                docs.append({'doc_id': doc['id'], 'url': url})
+
+            # Check if we need to go further
+            page_key = res.get('nextPageKey', None)    
+            if not page_key:  # Break the loop if there's no next page
+                break
+    
+        return docs
+
     def _index_file(self, filename: str, uri: str, metadata: Dict[str, Any]) -> bool:
         """
         Index a file on local file system by uploading it to the Vectara corpus.
--- a/crawlers/CRAWLERS.md
+++ b/crawlers/CRAWLERS.md
@@ -28,6 +28,7 @@ website_crawler:
     extraction: playwright
     keep_query_params: false
     crawl_report: false
+    remove_old_content: false
     ray_workers: 0
 ...
 ```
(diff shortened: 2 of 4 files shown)


adeelehsan (approved):

core/indexer.py
```diff
@@ -192,6 +192,47 @@ def delete_doc(self, doc_id: str) -> bool:
...
+        Returns:
+            list of [docId, URL] for each document in the corpus
+            URL taken from metadata if exists, otherwise it'd be None
+        """
+        page_key = None  # Initialize page_key as None
+        docs = []
+    
+        # Loop until there's no next page
+        while True:
+            body = {"corpusId": self.corpus_id, "numResults": 100}
```
justinhayes, line 207: The max page size is 1000. Why not use 1000 instead of 100 here? Larger batches should result in a faster overall time to cover all of the documents.

crawlers/CRAWLERS.md
```diff
@@ -45,7 +46,9 @@ Other parameters:
 - `pos_regex` defines one or more (optional) regex expressions defining URLs to match for inclusion.
 - `neg_regex` defines one or more (optional) regex expressions defining URLs to match for exclusion.
 - `keep_query_params`: if true, maintains the full URL including query params in the URL. If false, then it removes query params from collected URLs.
-- `crawl_report`: if true, creates a file under ~/tmp/mount called "crawl_report.txt" that lists all URLs crawled
+- `crawl_report`: if true, creates a file under ~/tmp/mount called `urls_indexed.txt` that lists all URLs crawled
+- `remove_old_content`: if true, removed any URL that currently exists in the corpus but is NOT in this crawl. CAUTION: this removed data from your corpus.
```
justinhayes, line 50: In the description of the remove_old_content option, change both occurrences of the word "removed" to "removes"

crawlers/CRAWLERS.md
```diff
@@ -184,7 +188,9 @@ It has two parameters
 - `doc_system` is a text string specifying the document system crawled, and is added to the metadata under "source"
 - `ray_workers` if it exists defines the number of ray workers to use for parallel processing. ray_workers=0 means dont use Ray. ray_workers=-1 means use all cores available.
 - `num_per_second` specifies the number of call per second when crawling the website, to allow rate-limiting. Defaults to 10.
-- `crawl_report`: if true, creates a file under ~/tmp/mount called "crawl_report.txt" that lists all URLs crawled
+- `crawl_report`: if true, creates a file under ~/tmp/mount called `urls_indexed.txt` that lists all URLs crawled
+- `remove_old_content`: if true, removed any URL that currently exists in the corpus but is NOT in this crawl. CAUTION: this removed data from your corpus.
```
justinhayes, line 192: Same as the other comment... replace "removed" with "removes"

crawlers/docs_crawler.py
```diff
@@ -134,10 +134,25 @@ def crawl(self) -> None:
...
+        else:
+            logging.info(f"Collected {len(self.crawled_urls)} URLs to crawl and index.")
+
+        # Determine which URLs to remove from corpus
+        if self.cfg.docs_crawler.get("remove_old_content", False):
+            existing_docs = self.indexer._list_docs()
+            docs_to_remove = [t for t in existing_docs if t['url'] and t['url'] not in self.crawled_urls]
+            for doc in docs_to_remove:
+                if doc['url']:
+                    self.indexer.delete_doc(doc['doc_id'])
```
justinhayes, line 150: I think it is safer to remove the old content after the crawl of the new content completes. If something goes wrong with crawling the new docs then I'd rather there still be the old docs in the corpus. Alternatively, if you delete all the old docs first then the crawl for the new docs fails you are left with a mostly empty corpus.

crawlers/website_crawler.py
```diff
@@ -99,17 +99,34 @@ def crawl(self) -> None:
...
+            logging.info(f"Collected {len(urls)} URLs to crawl and index.")
+
+        # Determine which URLs to remove from corpus
+        if self.cfg.website_crawler.get("remove_old_content", False):
+            existing_docs = self.indexer._list_docs()
+            logging.info(f"DEBUG existing_docs = {existing_docs}, urls = {urls}")
+            docs_to_remove = [t for t in existing_docs if t['url'] and t['url'] not in urls]
+            for doc in docs_to_remove:
+                if doc['url']:
+                    self.indexer.delete_doc(doc['doc_id'])
```
justinhayes, line 119: I think it is safer to remove the old content after the crawl of the new content completes. If something goes wrong with crawling the new docs then I'd rather there still be the old docs in the corpus. Alternatively, if you delete all the old docs first then the crawl for the new docs fails you are left with a mostly empty corpus.

core/indexer.py
ofermend: Good idea. Done

justinhayes:
Approved

PR #87, opened 2024-05-06
integrated ray with slack crawler

adeelehsan:


ofermend (approved):
LGTM. DIdn't test locally

PR #90, opened 2024-05-10
updates to make slack crawler better

ofermend:


ofermend:
Added correct logging with ray
split processing to get main messages locally but add threads msgs and index using Ray

adeelehsan (approved):

PR #94, opened 2024-05-18
Youtube crawler

ofermend:
Added Youtube playlist crawler
Removed some packages replaced by inline code


adeelehsan (approved):
Looks good. Did't test it locally.

PR #95, opened 2024-06-03
Improve edgar crawler

ofermend:
Refactor of Edgar (financial data) crawler
Simpilfication of html_to_text function to be more robust


justinhayes (approved):

adeelehsan (approved):
Looks good. 
I have tested few example with the unicodedata.normalize() and it seems to bi working fine. E.g I tried `unicodedata.normalize('NFC', '\u0061\u0301')` and output was `á`

PR #101, opened 2024-07-02
google drive crawler

AbhilashaLodha:
- Added a new config file `vectara-gdrive.yaml` for the google drive under the folder `config`.
- Added the python script for crawling the google-drive named `gdrive_crawler.py` under the folder `crawlers`
- Added the condition to extract delegated-users in `ingest.py`
- Added some more packages in `requirements.txt`


crawlers/gdrive_crawler.py
```diff
@@ -0,0 +1,276 @@
...
+    def __init__(self, cfg: OmegaConf, endpoint: str, customer_id: str, corpus_id: int, api_key: str, delegated_users: List[str]) -> None:
+        super().__init__(cfg, endpoint, customer_id, corpus_id, api_key)
+        logging.info("Google Drive Crawler initialized")
+
+        self.delegated_users = delegated_users
+        self.creds = None
+        self.service = None
+        self.processed_files = self.load_processed_files()
+
+    def load_processed_files(self):
```
ofermend, line 111: As discussed, this local storage is not a good enough solution for production deployment, as this is not always maintained (docker can fail and restart). Let's replace with the days_since solution and remove this.

crawlers/gdrive_crawler.py
```diff
@@ -0,0 +1,276 @@
...
+                        results.append(file)
+                page_token = response.get('nextPageToken', None)
+                if not page_token:
+                    break
+            except HttpError as error:
+                print(f"An error occurred: {error}")
+                break
+        return results
+
+    def handle_file(self, file):
```
ofermend, line 152: I see a few types of files are unhandled, e.g. PDF which is important.
And specifically for all files they are loaded, parsed locally into text, and then uploaded as text.
Why not use FILE_UPLOAD for the file types we support (like PDF, DOC, PPT, etc) and the method used here for all others?

crawlers/gdrive_crawler.py
AbhilashaLodha: Done! I have refactored the code to use date filtering and removed local storage for processed files.

crawlers/gdrive_crawler.py
```diff
@@ -0,0 +1,199 @@
...
+SCOPES = ["https://www.googleapis.com/auth/drive.readonly"]
+SERVICE_ACCOUNT_FILE = 'credentials.json'
+
+def get_credentials(delegated_user):
+    credentials = service_account.Credentials.from_service_account_file(
+        SERVICE_ACCOUNT_FILE, scopes=SCOPES)
+    delegated_credentials = credentials.with_subject(delegated_user)
+    return delegated_credentials
+
+def sanitize_filename(name):
```
ofermend, line 24: No need to implement this as the https://pypi.org/project/python-slugify/ package does exactly that. Maybe remove this function and call slugify() instead?

crawlers/gdrive_crawler.py
```diff
@@ -0,0 +1,199 @@
...
+
+def download_file(service, file_id):
+    try:
+        request = service.files().get_media(fileId=file_id)
+        file = io.BytesIO()
+        downloader = MediaIoBaseDownload(file, request)
+        done = False
+        while not done:
+            status, done = downloader.next_chunk()
+            print(f"Download {int(status.progress() * 100)}.")
```
ofermend, line 36: print() messages don't show with vectara-ingest. If you want this to show up in the logs, please use logging.info()

crawlers/gdrive_crawler.py
```diff
@@ -0,0 +1,199 @@
...
+    return delegated_credentials
+
+def sanitize_filename(name):
+    # Remove or replace characters that are not allowed in filenames
+    return "".join(c if c.isalnum() or c in (' ', '.', '_', '-') else '_' for c in name)
+
+def download_file(service, file_id):
+    try:
+        request = service.files().get_media(fileId=file_id)
+        file = io.BytesIO()
```
ofermend, line 31: This is actually not a file. It's just an in-memory bytestream, right?
so maybe change the name of the variable. 

There is also the risk that this may take too much memory for large files. But I think it's okay to note that for this time as a comment, and we can fix it sometime later.

crawlers/gdrive_crawler.py
```diff
@@ -0,0 +1,199 @@
...
+            modified_time = file.get('modifiedTime', 'N/A')
+            owners = ', '.join([owner['displayName'] for owner in file.get('owners', [])])
+            size = file.get('size', 'N/A')
+
+            print(f'\nCrawling file {name}')
+
+            file_metadata = {
+                'id': file_id,
+                'name': name,
+                'created_at': created_time,
```
ofermend, line 172: Are these (created_time and modified_time) strings or "datetime" type?

crawlers/gdrive_crawler.py
```diff
@@ -0,0 +1,199 @@
...
+        date_threshold = (datetime.utcnow() - timedelta(days=N)).isoformat() + 'Z'
+        
+        for user in self.delegated_users:
+            print(f"Processing files for user: {user}")
+            self.creds = get_credentials(user)
+            self.service = build("drive", "v3", credentials=self.creds)
+            
+            list_files = self.list_files(self.service, date_threshold=date_threshold)
+            for file in list_files:
+                modified_time = file.get('modifiedTime', 'N/A')
```
ofermend, line 197: So if there is no modifiedTime, it would return a string with the value "N/A" and then do string comparison against date_threshold which is a string. So that seems wrong. 
ALso - if modifiedTime exists and is returned properly from the file object, is it the same format as date_threshold? (which looks something like this: '2024-07-10T01:23:03.556652Z')?

crawlers/gdrive_crawler.py
```diff
@@ -187,13 +183,22 @@ def crawl(self) -> None:
...
+        
         for user in self.delegated_users:
-            print(f"Processing files for user: {user}")
+            logging.info(f"Processing files for user: {user}")
             self.creds = get_credentials(user)
             self.service = build("drive", "v3", credentials=self.creds)
             
             list_files = self.list_files(self.service, date_threshold=date_threshold)
             for file in list_files:
                 modified_time = file.get('modifiedTime', 'N/A')
```
ofermend, line 197: Sorry I don't understnad this logic
1. If modifiedTime does not exist, did you intend for the file to not be crawled? Isn't there a way to still crawl the file instead of ignoring it?
2. Generally in Python it's best practice to use "None" (not the string, the value). So perhaps you can do "modified_time = file.get('modifiedTime', None) and use that subsequently?
3. Generally I would use the datetime functionality (instead of strings) in Python for the thresholding by date. See for exampe: https://github.com/vectara/vectara-ingest/blob/main/crawlers/rss_crawler.py#L20
Working with strings is not the preferred way, and can have multiple ways it may go wrong given the complexity of dates, timezones, formatting, etc

requirements.txt
```diff
@@ -19,7 +19,8 @@ biopython==1.81
 boto3==1.26.116
 mwviews==0.2.1
 toml==0.10.2
-pandas==1.3.5
+pandas==2.2.2
```
ofermend, line 22: Why do we need to upgrade pandas and numpy?
This may break other things, so unless it's tested thoroughly I would avoid without a specific need.

requirements.txt
```diff
@@ -52,4 +53,7 @@ pydub==0.25.1
 pytube==15.0.0
 openai-whisper==20231117
 youtube-transcript-api==0.6.2
-sec-downloader==0.11.1
\ No newline at end of file
+sec-downloader==0.11.1
+python-docx==1.1.2
```
ofermend, line 57: Are these 3 still needed? (python-docx, openpyxl, and python-pptx)

ingest.py
```diff
@@ -129,15 +133,30 @@ def main() -> None:
...
+            Crawler, 'crawlers', f'{crawler_type.capitalize()}Crawler',
+            cfg, endpoint, customer_id, corpus_id, api_key, delegated_users
+        )
+    else:
+        crawler = instantiate_crawler(
+            Crawler, 'crawlers', f'{crawler_type.capitalize()}Crawler',
+            cfg, endpoint, customer_id, corpus_id, api_key
+        )
 
+    logging.info(f"Crawing instantiated...")
```
ofermend, line 159: Typo: should be "crawling"

crawlers/gdrive_crawler.py
```diff
@@ -0,0 +1,204 @@
...
+    def crawl(self) -> None:
+        N = 7  # Number of days to look back
+        date_threshold = datetime.utcnow() - timedelta(days=N)
+        
+        for user in self.delegated_users:
+            logging.info(f"Processing files for user: {user}")
+            self.creds = get_credentials(user)
+            self.service = build("drive", "v3", credentials=self.creds)
+            
+            list_files = self.list_files(self.service, date_threshold=date_threshold.isoformat() + 'Z')
```
ofermend, line 191: I see  you using the ISO format + "z": is this because Google API demands it to be in UTC format and in this way?

crawlers/gdrive_crawler.py
```diff
@@ -0,0 +1,204 @@
...
+            for file in list_files:
+                modified_time = file.get('modifiedTime', None)
```
ofermend, line 193: This is unneeded. The date_threshold check was already done in list_files() so here all files should already be in the right date range.

Dockerfile
```diff
@@ -34,6 +34,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 # install python packages
 WORKDIR ${HOME}
 COPY requirements.txt requirements-extra.txt $HOME/
+COPY crawlers/credentials.json $HOME/
```
ofermend, line 37: This is not a good idea. I'll explain offline why.

ofermend (approved):
LGTM

PR #107, opened 2024-07-24
updated notion crawler

ofermend:


ofermend:
Added the option to remove old content from notion
Also not including blocks that are sub-pages to avoid duplicate content.

PR #114, opened 2024-08-21
Update website_crawler.py to keep clean urls

nespera:
It looks like the website crawler normalizes the URLs it gathers but then immediately throws them away. This change keeps the normalized list.

### DIFF
--- a/crawlers/website_crawler.py
+++ b/crawlers/website_crawler.py
@@ -79,7 +79,6 @@ class WebsiteCrawler(Crawler):
                                            pos_regex=self.pos_regex, neg_regex=self.neg_regex, 
                                            indexer=self.indexer, visited=set(), verbose=self.indexer.verbose)
                 urls = clean_urls(urls_set, keep_query_params)
-                urls = list(set(urls_set))
             else:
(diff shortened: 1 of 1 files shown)


crawlers/website_crawler.py
```diff
@@ -79,7 +79,6 @@ def crawl(self) -> None:
                                            pos_regex=self.pos_regex, neg_regex=self.neg_regex, 
                                            indexer=self.indexer, visited=set(), verbose=self.indexer.verbose)
                 urls = clean_urls(urls_set, keep_query_params)
-                urls = list(set(urls_set))
```
ofermend, line 82: good catch, but I think the idea of line 82 is really to remove duplicates, so should have been
"urls = list(set(urls))"
instead of removing this altogether.

nespera: But clean_urls is also removing duplicates with the same mechanism isn't it?

ofermend: yes you are right!

ofermend (approved):
LGTM.

PR #115, opened 2024-08-29
Only normalize metadata values that are strings

nespera:
The values in the dictionary are declared to be Any but the normalization code assumes strings, so only normalize if they really are strings.


ofermend (approved):
Thanks for the fix. LGTM

PR #119, opened 2024-09-14
Store docs locally

ofermend:
fixed issue with detect_file_type
added store_docs configuration option to store all file content locally

### DIFF
--- a/core/utils.py
+++ b/core/utils.py
@@ -4,6 +4,7 @@ from urllib.parse import urlparse, urlunparse, ParseResult
 from pathlib import Path
 
 from bs4 import BeautifulSoup
+import xml.etree.ElementTree as ET
 from urllib.parse import urljoin, urlparse
 from slugify import slugify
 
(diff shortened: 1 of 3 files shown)


core/utils.py
```diff
@@ -52,10 +60,11 @@ def detect_file_type(file_path):
...
     mime = magic.Magic(mime=True)
     mime_type = mime.from_file(file_path)
-    with open(file_path, 'r', encoding='utf-8') as file:
-        first_1024_bytes = file.read(1024)
-    if '<html' in first_1024_bytes.lower() and '</html>' in first_1024_bytes.lower():
-        return 'text/html'
+    if mime_type in ['text/html', 'application/xml', 'text/xml']:
+        with open(file_path, 'r', encoding='utf-8') as file:
+            first_1024_bytes = file.read(1024)
+        if '<html' in first_1024_bytes.lower() and '</html>' in first_1024_bytes.lower():
```
justinhayes, line 66: Why is this looking for the closing html tag in the first 1024 bytes? Shouldn't that be looked for in the last 1024 bytes?

ofermend: It is actually okay since usually XML files look this way (very close start/end HTML tags), but let me see if I can figure out a more robust detection option.

PR #131, opened 2024-12-02
Chunking update

ofermend:
1. Added chunking with unstructured
2. Refactored indexing so that process of local file is in a new module
3. Other improvements and bug fixes
4. Update to requirements.txt

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
+RUN rm -f /usr/share/fonts/truetype/unifont/unifont_sample.ttf /usr/share/fonts/truetype/noto/NotoColorEmoji.ttf
 ENV OMP_NUM_THREADS=4
 
 # Install python packages
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ vectara:
   # If using Unstructured to process files locally, we define a few arguments
   unst_chunking_strategy: none            # chunking strategy to use: basic, by_title or none; default none
   unst_chunk_size: 1024                   # chunk size if using unstructured chunking; default 1024
-  unst_use_core_indexing: true            # whether to use core_indexing, or let Vectara chunk further
+  unst_use_core_indexing: true            # whether to use core_indexing which maintains the chunks from unstructured, or let vectara chunk further
 
   # Whether masking of PII is attempted on all text fields (title, text, metadata values)
   # Notes: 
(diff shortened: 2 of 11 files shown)


Dockerfile
```diff
@@ -31,6 +31,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
+RUN rm -f /usr/share/fonts/truetype/unifont/unifont_sample.ttf /usr/share/fonts/truetype/noto/NotoColorEmoji.ttf
```
ofermend, line 34: prevents warning msg on missing fonts

README.md
```diff
@@ -197,17 +197,24 @@ vectara:
...
+  #    You can modify this to use one of the alternatives: https://unstructured-io.github.io/unstructured/best_practices/models.html) if you want a slower but more performance model.
   # See [here](TABLE_SUMMARY.md) for some examples of how table summary works.
   summarize_tables: false
-  
+  summarize_images: false
+
+  # If using Unstructured to process files locally, we define a few arguments
+  unst_chunking_strategy: none            # chunking strategy to use: basic, by_title or none; default none
+  unst_chunk_size: 1024                   # chunk size if using unstructured chunking; default 1024
+  unst_use_core_indexing: true            # whether to use core_indexing, or let Vectara chunk further
```
justinhayes, line 216: I suggest using the following for the comment for unst_use_core_indexing: "# whether to use core_indexing which maintains the chunks from unstructured, or let vectara chunk further". Or something like that, just to make it clearer on exactly what happens here.

core/doc_parser.py
```diff
@@ -0,0 +1,141 @@
...
+                if inx>0 and type(elements[inx-1]) in [us.documents.elements.Title, us.documents.elements.NarrativeText]:
+                    image_summary = self.image_summarizer.summarize_image(e.metadata.image_path, elements[inx-1].text)
+                else:
+                    image_summary = self.image_summarizer.summarize_image(e.metadata.image_path, None)
+                if image_summary:
+                    texts.append(image_summary)
+
+        # No chunking strategy may result in title elements; if so - use the first one as doc_title
+        titles = [str(x) for x in elements if type(x)==us.documents.elements.Title and len(str(x))>10]
+        doc_title = titles[0] if len(titles)>0 else 'no title'
```
justinhayes, line 138: Should we set doc_title to none or the empty string here, instead of 'no title'? I didn't look at all downstream code, but is there a risk of the text 'no title' being indexed as the actual title of the document?

justinhayes (approved):
I made a few comments. So once you check those out and address them one way or another, then this all looks good to me.

joaosal (approved):

PR #134, opened 2024-12-10
Updates to parsing

ofermend:
- Supports docling and Unst parsing for HTML file (in index_url)
- Added ability to define metadata fields that are extracted using LLM from the document

### DIFF
--- a/core/doc_parser.py
+++ b/core/doc_parser.py
@@ -2,7 +2,8 @@ import logging
 from typing import List, Tuple
 import time
 
-from core.utils import TableSummarizer, ImageSummarizer, detect_file_type
+from core.summary import TableSummarizer, ImageSummarizer
+from core.utils import detect_file_type
 
 import unstructured as us
 from unstructured.partition.pdf import partition_pdf
@@ -56,6 +57,7 @@ class DoclingDocumentParser(DocumentParser):
     def parse(
             self,
             filename: str, 
+            source_url: str = "No URL"
         ) -> Tuple[str, List[str], List[dict]]:
         """
         Parse a local file and return the title and text content.
@@ -109,7 +111,7 @@ class DoclingDocumentParser(DocumentParser):
                 if image:
                     with open(image_path, 'wb') as fp:
                         image.save(fp, 'PNG')
-                    image_summary = self.image_summarizer.summarize_image(image_path, None)
+                    image_summary = self.image_summarizer.summarize_image(image_path, source_url, None)
                     if image_summary:
                         texts.append(image_summary)
                         metadatas.append({'parser_element_type': 'image'})
@@ -192,6 +194,7 @@ class UnstructuredDocumentParser(DocumentParser):
     def parse(
             self,
             filename: str, 
+            source_url: str = "No URL",
         ) -> Tuple[str, List[str], List[dict]]:
         """
         Parse a local file and return the title and text content.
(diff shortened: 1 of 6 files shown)


core/indexer.py
```diff
@@ -235,12 +270,33 @@ def fetch_page_contents(self, url: str, remove_code: bool = False, debug: bool =
...
+            'text': text, 
+            'html': html, 
+            'title': title,
+            'url': out_url, 
+            'links': links,
+            'images': images,
+            'tables': tables
         }
 
+    def _scroll_to_bottom(self, page, pause=2000):
```
ofermend, line 284: This is to support html pages where the content only shows up once you scroll to the bottom.

core/doc_parser.py
```diff
@@ -56,6 +57,7 @@ def _lazy_load_docling():
     def parse(
             self,
             filename: str, 
+            source_url: str = "No URL"
```
justinhayes, line 60: Curious as to the logic behind setting source_url to "No URL" in this case, instead of setting it to the empty string. Does it have to have some value for downstream code to work? In general I think it's better to set something to empty if it is not known, in the event that some downstream code checks to see if the value is set to something non-empty, and if it is then it assumes it is a value URL (which in this case it would not be).

core/summary.py
```diff
@@ -0,0 +1,141 @@
...
+from openai import OpenAI
+
+def get_attributes_from_text(text: str, metadata_questions: list[str], openai_api_key) -> Set[str]:
+    """
+    Given a text string, ask GPT-4o to answer a set of questions from the text
+    Returns a dictionary of question/answer pairs.
+    """
+    prompt = f"""
+        Here is text: {text}.
+        Your task is to answer each of the following questions, based on the text:
```
justinhayes, line 17: In the first line of the prompt you mention "text" in the context of the source data to be looked at. Then in this line you mention "text" in the context of the specific questions to be answered. Will the LLM get confused about multiple uses of the term "text" to refer to different things? Perhaps consider not using "text" in this line?

core/summary.py
```diff
@@ -0,0 +1,141 @@
...
+        content = None
+        with open(image_path, "rb") as f:
+            content = base64.b64encode(f.read()).decode("utf-8")
+        
+        width, height = get_image_shape(content)
+        if width<10 or height<10:
+            logging.info(f"Image too small to summarize ({image_url})")
+            return None
+
+        prompt = """
```
justinhayes, line 59: In the scenario where the image contains lots of text (e.g. engineering schematics with labels) this could look more like an OCR operation than an image summarization operation. I wonder if it would help to add something to the prompt to talk about specifically extracting any text strings that are shown within the image? The prompt as-is might already do that (e.g. "what information they convey") but I would for sure test this on images that have a lot of text that a user might want to retrieve.

core/summary.py
```diff
@@ -0,0 +1,141 @@
...
+        except Exception as e:
+            logging.info(f"Failed to summarize image ({image_url}): {e}")
+            return None
+
+class TableSummarizer():
+    def __init__(self, openai_api_key: str):
+        self.client = OpenAI(api_key=openai_api_key)
+
+    def summarize_table_text(self, text: str):
+        prompt = f"""
```
justinhayes, line 104: Does this handle the use case where someone wants simply to look up information from within the table (and not actually ask about a summary of the broader table)? E.g. if the rows represent companies and the columns represent financial metrics (revenue, debt, etc), and the user simply wants to ask "what is the revenue for Capital One?" then will this prompt extract the values for all the different columns for each row in such a way that the context for each column is preserved?

This might be out of scope for the purpose of this table summarization operation, but I wanted to see if this use case is covered.

justinhayes (approved):
approved once the comments are addressed

sulekz (approved):

core/summary.py
ofermend: No - this is actually the same text. I'm asking the LLM to answer the list of questions based on the text provided (which is the full document)

core/summary.py
ofermend: Good idea. I don't think it hurts to add specific instructions for this.

core/summary.py
ofermend: This was out of scope, and my concern was that it may become too big of a summary. But it may also be okay - just did not test it. I'll put a note to try that later on.

PR #138, opened 2024-12-17
Upload with APIv2

ofermend:
Enable file_upload with APIv2 using data table understanding


sulekz (approved):
Are you planning to deprecate v1 indexing? This would eliminate the need for the `corpus_key` as well as `corpus_id` to be passed to the crawler.

ofermend:
Soon. Planning a major upgrade to APIv2 (Shane already started on it), but this would be mid-way step.

PR #153, opened 2025-02-05
Removed hard coded path for the path in the config. Fixes #152

### LINKED ISSUE #152: folder crawler has hard coded path.

jcustenborder:


crawlers/folder_crawler.py
```diff
@@ -41,7 +41,7 @@ def process(self, file_path: str, file_name: str, metadata: dict):
 class FolderCrawler(Crawler):
 
     def crawl(self) -> None:
-        folder = "/home/vectara/data"
+        folder =  self.cfg.folder_crawler.path
```
ofermend, line 44: This requires every configuration file to explicitly specify this. I suggest that you put in the default for backwards compatibility, so: folder = self.cfg.folder_crawler.get("path", "/home/vectara/data"). Note also that if used within Docker and providing a non-existent folder - it may not work, so we need to add some checks that this folder. It is also referred to in run.sh

ofermend:
@jcustenborder - the "path" argument under "folder_crawler" is an existing argument and in run.sh it's mapped to /home/vectara/data. But making this change it will fail in Docker so I don't think this works. Can u pls change the code so that it works in both the DOcker and non-docker settings?

jcustenborder:
This should be corrected.

PR #173, opened 2025-02-18
Update to last_updated

ofermend:
updated extraction of last_updated with docs_crawler


core/indexer.py
```diff
@@ -21,6 +21,10 @@
 import markdown
 import whisper
 
+from bs4 import BeautifulSoup
+from email.utils import parsedate_to_datetime
+import hashlib
```
adeelehsan, line 26: Just a nit: it should be above the from

adeelehsan (approved):

PR #176, opened 2025-02-21
PDF Chunking support

jcustenborder:


core/indexer.py
```diff
@@ -978,8 +981,36 @@ def index_file(self, filename: str, uri: str, metadata: Dict[str, Any], id: str
                 ex_metadata = {}
             metadata.update(ex_metadata)
 
-            # index the file within Vectara (use FILE UPLOAD API)
-            succeeded = self._index_file(filename, uri, metadata, id)
+            max_pdf_size = 50
```
ofermend, line 984: Why not use size_limit from above? (maybe max_pdf_size is a better name, but could use a single variable for both purposes)?

core/indexer.py
```diff
@@ -978,8 +981,36 @@ def index_file(self, filename: str, uri: str, metadata: Dict[str, Any], id: str
...
+            if filesize_mb > max_pdf_size:
+                pdf_reader = PdfReader(filename)
+                total_pages = len(pdf_reader.pages)
+                logging.info(f"{filename} is {filesize_mb} which is larger than {max_pdf_size} mb with #{total_pages} pages. Splitting into #{pages_per_pdf} chunks")
+                error_count = 0
+                for i in range(0, total_pages, pages_per_pdf):
+                    pdf_writer = PdfWriter()
+                    for j in range(i, min(i + pages_per_pdf, total_pages)):
+                        pdf_writer.add_page(pdf_reader.pages[j])
+                    part_id = f"#{filename}-#{i}"
```
ofermend, line 996: Can we maintain the same schema for document ID as is used in _index_file?
There we use filename.split('/')[-1], so here it would be something like f"{filename.split('/')[-1]-{i}" perhaps?

core/indexer.py
```diff
@@ -977,9 +979,42 @@ def index_file(self, filename: str, uri: str, metadata: Dict[str, Any], id: str
             else:
                 ex_metadata = {}
             metadata.update(ex_metadata)
-
-            # index the file within Vectara (use FILE UPLOAD API)
-            succeeded = self._index_file(filename, uri, metadata, id)
+            metadata['file_name'] = filename.split('/')[-1]
+
+            max_pdf_size = int(self.cfg.vectara.get('max_pdf_size', 50))
```
ofermend, line 984: Add these in README

core/indexer.py
```diff
@@ -977,9 +979,42 @@ def index_file(self, filename: str, uri: str, metadata: Dict[str, Any], id: str
...
+                error_count = 0
+                for i in range(0, total_pages, pages_per_pdf):
+                    pdf_writer = PdfWriter()
+                    part_metadata = {
+                        "start_page": i,
+                        "end_page": min(i + pages_per_pdf, total_pages)
+                    }
+                    for j in range(i, min(i + pages_per_pdf, total_pages)):
+                        pdf_writer.add_page(pdf_reader.pages[j])
+                    part_id = f"{metadata['file_name']}-{i}"
```
ofermend, line 1000: I find the name "part_id" and "part_metadata" a bit confusing since in the context of Vectara, a "part" has another meaning. Maybe we can call this "pdf_part_id" and "pdf_part_metadata" or something like that?

ofermend (approved):
I made a few minor comments, approved after those are taken care of.

ofermend (approved):
LGTM

PR #179, opened 2025-02-24
External SSL Certificate support

jcustenborder:
For this one I have things configured so that you can have different certificate authorities for different things. For example I can see a customer connecting to ServiceNow which is cloud hosted but having an onprem instance of Vectara. In this case you most likely would want to use the default certificates for ServiceNow but use an internal certificate authority for the onprem instance of Vectara. This PR allows you to configure both.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,10 @@ vectara:
   # Which whisper model to use for audio files (relevant for YT, S3 and folder crawlers)
   # Valid values: tiny, base, small, medium or large. Defaults to base.
   whisper_model: the model name for whisper
-
+  # Whether the session should trust the environment settings. When set to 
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -204,7 +204,10 @@ vectara:
   # Which whisper model to use for audio files (relevant for YT, S3 and folder crawlers)
   # Valid values: tiny, base, small, medium or large. Defaults to base.
   whisper_model: the model name for whisper
-
+  # Whether the session should trust the environment settings. When set to False SSL will not be verified. Do not use in production.
+  ssl_trust_env: None
```
ofermend, line 208: Set to False or True

ofermend (approved):
LGTM

PR #188, opened 2025-03-13
Sharepoint

jcustenborder:


crawlers/sharepoint_crawler.py
```diff
@@ -0,0 +1,162 @@
...
+class SharepointCrawler(Crawler):
+    """
+    A crawler implementation for ingesting and indexing documents from SharePoint sites.
+
+    The SharepointCrawler class connects to SharePoint, authenticates using specified credentials,
+    recursively crawls specified folders, downloads supported document types, and indexes them
+    using the configured indexing service.
+    """
+
+    def new_url(self, /, *paths) -> furl:
```
nikron, line 21: not sure what the / does, doesn't seem useful?

crawlers/sharepoint_crawler.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        Generates a direct download URL for the given SharePoint file.
+
+        Args:
+            file: A SharePoint file object with a server-relative URL attribute.
+
+        Returns:
+            str: Direct URL to download the specified file.
+        """
+        download_url = self.new_url("_layouts/15/download.aspx")
```
nikron, line 46: would be helpful to explain this harcoded string

crawlers/sharepoint_crawler.py
```diff
@@ -0,0 +1,162 @@
...
+            with tempfile.NamedTemporaryFile(suffix=file_extension, mode="wb", delete=False) as f:
+                logging.debug(f"Downloading and writing content for {file.unique_id} to {f.name}")
+                file.download_session(f).execute_query()
+                f.flush()
+                f.close()
+
+                try:
+                    succeeded = self.indexer.index_file(f.name, metadata['url'], metadata, file.unique_id)
+                finally:
+                    if os.path.exists(f.name):
```
nikron, line 136: why is delete=False if you just delete here?

nikron (approved):

crawlers/sharepoint_crawler.py
jcustenborder: We use this pattern in other places. Basically it allows you to set a baseurl. In this case `https://somesite.sharepoint.com/subdirectory` then request subsequent pages from it. So for example in the other call `self.new_url("_layouts/15/download.aspx")` would give you a furl object that pointing to `https://somesite.sharepoint.com/subdirectory/_layouts/15/download.aspx` that I can later set the query string on with a dictionary like interface.

crawlers/sharepoint_crawler.py
jcustenborder: LOL that is sharepoint. That is the url for downloading content.

crawlers/sharepoint_crawler.py
jcustenborder: tempfile.NamedTemporaryFile will delete the file on close unless you do this. I ran into a file locking issue in another crawler where just a flush() wouldn't work. In this case we close the file but don't allow tempfile to delete it, then delete it after an upload is successful.

crawlers/sharepoint_crawler.py
nikron: yeah i figured, i'm just saying leaving a comment linking to the docs or explaining where it comes from would be useful

PR #194, opened 2025-03-28
Url configuration validation and debug logging

jcustenborder:


core/utils.py
```diff
@@ -40,10 +42,14 @@
 binary_extensions = archive_extensions + img_extensions + doc_extensions
 
 def setup_logging():
+    log_level_str = os.getenv("LOGGING_LEVEL", "INFO").upper()
```
ofermend, line 45: Would this work as it's running in Docker? doesn't it need to be in secrets.toml to work? or maybe just do it as a configuration param.

jcustenborder: I added support for it to [run.sh](https://github.com/vectara/vectara-ingest/pull/194/files#diff-d31ce0453051853c17ba2a5225b3d1bfab548e095bab0967d6acfd1b3ce1b35dR80)

`LOGGING_LEVEL=debug ./run.sh config-private/vectara-sharepoint.yaml sharepoint`

nikron (approved):

ingest.py
```diff
@@ -241,8 +248,21 @@ def main() -> None:
...
+    if not is_valid_url(api_url):
+        raise Exception(f"endpoint '{api_url}' could not be parsed to a valid URL.")
+
+    auth_url = cfg.vectara.get("auth_url", "https://auth.vectara.io")
+    if not auth_url.startswith(("http://", "https://")):
+        logging.warning(f"Correcting auth_url {auth_url} to https://{auth_url}. This will error in a future release.")
+        auth_url = f"https://{auth_url}"
+    if not is_valid_url(auth_url):
+        raise Exception(f"endpoint '{auth_url}' could not be parsed to a valid URL.")
+
```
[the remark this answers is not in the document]
jcustenborder: Isn't that what I am doing? I'm checking for `http://` or `https://`, prepending it if it does not exist, then checking for a valid url. I did it this way because I wanted to be able to set `endpoint: http://localhost:8080` for local testing for example.

PR #208, opened 2025-05-12
bugfix

ofermend:


jcustenborder (approved):
LGTM. Can we swap the logging level to warn?

PR #209, opened 2025-05-12
adding setup to run it using cli

adeelehsan:


setup.sh
```diff
@@ -0,0 +1,140 @@
+#!/bin/bash
```
adeelehsan, line 1: @ofermend should I keep it? If we want to keep it then I can update it use the conda.

ingest.py
```diff
@@ -260,42 +256,73 @@ def main() -> None:
...
-    secrets_path = os.environ.get('VECTARA_SECRETS_PATH', '/home/vectara/env/secrets.toml')
+    # Certificate auto-detection only in CLI mode (Docker has its own mechanism)
+    if not is_running_in_docker():
+        # If ssl_verify is not set, check for ca.pem or ssl/ directory
+        if not cfg.vectara.get("ssl_verify", None):
+            if os.path.exists("ca.pem"):
+                logger.info("Found ca.pem in current directory, using it for SSL verification")
+                OmegaConf.update(cfg, 'vectara.ssl_verify', os.path.abspath("ca.pem"))
+            elif os.path.isdir("ssl"):
+                logger.info("Found ssl/ directory in current directory, using it for certificates")
```
jcustenborder: Lets not do this in the CLI. This was for the docker container to use system certs. The user should update their system certs which would make this unnecessary.

jcustenborder:
We should not be doing if/else logic around if the application is running in docker. We should create configuration settings for all file paths and allow those configs to be exposed on the command line. For example:

The default in the code is the same as today. `/home/vectara/env/credentials.json`

We could add a method that builds command line parser based on a mapping. Something like this. This needs to be cleaned up. I just used ChatGPT to generate an example of what I'm proposing. 

```python
import argparse
from omegaconf import OmegaConf

mappings = {
    'google-credentials-path': 'google.credentials_path',
    'output-dir': 'vectara.output_dir',
}

def parse_command_line_and_update_config(cfg):
    parser = argparse.ArgumentParser()
    parser.add_argument("job_config_file")
    parser.add_argument("secrets_profile")

    for switch in mappings:
        parser.add_argument(f"--{switch}", dest=switch.replace('-', '_'))

    args = parser.parse_args()

    for switch, config_path in mappings.items():
        attr = switch.replace('-', '_')
        value = getattr(args, attr)
        if value is not None:
            OmegaConf.update(cfg, config_path, value)

    return args.job_config_file, args.secrets_profile
```

ofermend (approved):
LGTM!

PR #218, opened 2025-05-16
Add vectara fixed chunking + scrapy

ofermend:
1. Added an option to use Vectara's "fixed chunk size" during ingest
2. Added Scrapy crawler as an option for website crawler and docs crawler
3. Some bug fixes in docs crawler


jcustenborder (approved):
LGTM

core/crawler.py
```diff
@@ -1,77 +1,5 @@
...
-                    and     (len(neg_regex)==0 or (not any([r.match(u) for r in neg_regex]))) 
-                   ]
-        new_urls = list(set(new_urls))
-        new_urls = [u for u in new_urls if not any([u.endswith(ext) for ext in archive_extensions + img_extensions])]
-        visited.update(new_urls)
-
-        if len(new_urls) > 0:
-            logging.info(f"collected {len(visited)} URLs so far")
-            if verbose:
-                print(f"URLs so far: {visited}")
```
jcustenborder, line 65: Should we be using a print statement? Do we need to do verbose vs just using logging.debug? logging.debug you just run `LOGGING_LEVEL=debug ./run.sh`

Issue #219, opened 2025-05-21
Enabling doc_processing.do_ocr causes model to be downloaded every time.


jcustenborder:
This should be configurable and ideally should only happen once when the docker container is built.

PR #221, opened 2025-05-29
Download EasyOCR models during docker build.

jcustenborder:
This PR adds support for to download the EasyOCR models as part of the docker build process. When the buildarg `DOWNLOAD_EASYOCR_MODELS=true` the docker container will execute a script that will download all of the potential models to the container. This will help facilitate using EasyOCR in an airgapped environment. 

As part of this PR, all of the configuration settings for EasyOCR are now exposed under `doc_processing.easy_ocr_config`. By default the  model_storage_directory is configured to `os.path.expanduser("~/.EasyOCR/")` which is the same as the libraries default. Doing this ensures that the download during docker build and when the container is used are the same.

Added create_default_config to make default configuration values more readable. Defaults can be added here and the job config is merged over this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,6 +75,15 @@ RUN playwright install --with-deps firefox \
 # Set working directory
 WORKDIR ${HOME}
 
+ADD docker/bin/download-easyocr-models.py /bin/
+
+ARG DOWNLOAD_EASYOCR_MODELS=false
+
+RUN if [ "$DOWNLOAD_EASYOCR_MODELS" = "true" ]; then \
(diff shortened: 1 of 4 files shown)


jcustenborder:
Fixes #219

ofermend:
@jcustenborder - added copilot for review and pls see its comments
@adeelehsan - can u also pls take a look and make sure this does not conflict (or is otherwise accounted for to work well) in the conda/CLI version?

adeelehsan:
@jcustenborder is this ready to merge ?

adeelehsan (approved):

PR #222, opened 2025-05-30
Log msgs Docker command in red.

ofermend:
Per suggestion from user, to make sure it's clear what the container name is.


jcustenborder (approved):
LGTM

PR #226, opened 2025-06-04
Static metadata

jcustenborder:


core/indexer.py
```diff
@@ -708,6 +712,17 @@ def index_document(self, document: Dict[str, Any], use_core_indexing: bool = Fal
                 self.logger.info(f"Document {document['id']} already exists, skipping")
                 return False
 
+
+        if self.static_metadata:
```
ofermend, line 716: What about applying the same for non-file indexing?

justinhayes (commented):
Code looks good to me. But can you also update the main readme file to explain what this 'metadata' config is used for, which level it is set at, and what the value needs to look like?

PR #231, opened 2025-06-09
Upd csv crawler

ofermend:
TO upload a table as single thing.


ofermend:
Implemented in https://github.com/vectara/vectara-ingest/pull/266

PR #233, opened 2025-06-12
adding linux config and readme instruction

adeelehsan:


conda/README.md
```diff
@@ -37,6 +37,58 @@ After installing vectara-ingest, you can run it using the command-line interface
...
+# Fonts
+sudo yum install -y liberation-fonts dejavu-sans-fonts
+
+# Additional dependencies if available
+sudo yum install -y libcanberra-gtk3 atk || true
+```
+
+3. Install Firefox Browser
+
+If you need only Firefox (rather than all browsers that come with `playwright install`):
```
ofermend: I think for vectara-ingest you do only need firefox. So asking here may be confusing. Maybe we just say "for vectara-ingest you need to install firefox at a minimum" and then provide the instructions below. WDYT?

ofermend (approved):
LGTM. Did not test locally. One minor commeny.

PR #234, opened 2025-06-12
Generic CSV Crawling

jcustenborder:


core/dataframe_parser.py
```diff
@@ -0,0 +1,188 @@
...
+                table_dict = {'headers': cols, 'rows': rows, 'summary': table_summary}
+                return table_summary, table_dict
+            else:
+                return None
+        else:
+            logger.warning(f"Table summarization failed for the table with {df.shape[0]} rows and {df.shape[1]} columns.")
+            return None
+
+
+    def parse_element_dataframe(self, df:pd.DataFrame):
```
ofermend: This needs to be implemented, no?

ofermend: Based on the original code

core/dataframe_parser.py
```diff
@@ -0,0 +1,188 @@
+import logging
```
ofermend, line 1: Can u pls add doc_strings so it's easy for me and Adeel to understand the design and what each method is supposed to do?

core/dataframe_parser.py
```diff
@@ -0,0 +1,537 @@
...
+from omegaconf import DictConfig
+from core.summary import TableSummarizer
+from core.utils import html_table_to_header_and_rows
+from core.indexer import Indexer
+import unicodedata
+
+logger = logging.getLogger(__name__)
+
+
+class DataFrameMetadata(object):
```
ofermend, line 13: Why is sheet_names part of the top level base class? If we need sheet names for the XLS dataframe metadata - wouldn't this be only needed for that derived class?

core/dataframe_parser.py
```diff
@@ -0,0 +1,537 @@
...
+    else:
+        logger.debug(
+            f"load_dataframe_metadata('{file_path}) - using caller supplied data_frame_type='{data_frame_type}'"
+        )
+
+    logger.debug(
+        f"load_dataframe_metadata('{file_path}) - Loading as '{data_frame_type}'"
+    )
+
+    if "csv" == data_frame_type:
```
ofermend, line 298: Given that we have here an object model that differentiates the types (CSV, XLS, etc) - why not remove the if/then/else and use that to process in the derived class? (here and in other places below)

core/indexer.py
```diff
@@ -46,6 +47,28 @@
 warnings.filterwarnings("ignore", category=FutureWarning, module="whisper")
 warnings.filterwarnings("ignore", category=UserWarning, message="FP16 is not supported on CPU; using FP32 instead")
 
+get_headers = {
```
ofermend: Why override the get_headers function that's imported?

core/utils.py
```diff
@@ -350,13 +373,58 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         with self.lock:
             self.condition.notify()
 
+def get_urls_from_sitemap(homepage_url):
```
ofermend, line 376: Is this outdated e.g. relative to https://github.com/vectara/vectara-ingest/pull/235/files#diff-55b6e5d55d63395596fc7d855f242ae54496dd953576332c813c931ad4644b6eR392

ofermend (changes requested):
I provided some feedback in the PR. Please review

core/dataframe_parser.py
jcustenborder: That is not necessary. I can remove it.

core/dataframe_parser.py
jcustenborder: The Metadata classes are responsible for creating the dataframes so that the generic processing code can handle it. I don't see why this needs to change. We need to understand what type of dataframe we need to load the file as.

core/indexer.py
jcustenborder: Corrected

core/utils.py
jcustenborder: Can you help me understand this one? When I look the only changes I see are some logging and whitespace changes.

core/dataframe_parser.py
ofermend: It's not that it needs to change but it's just that the class hierarchy here is defined but not properly used. The idea would be that since you know which type it is given its class you don't need to do if/then/else. Happy to explain live if needed.

ofermend (changes requested):
Approved with plan for follow up PR to fix the following:
1. Simplify class hierarchy
2. Make load_dataframe (renamed load_dataframes) be able to load one or more so it generalizes to CSV cvs XLSX which may have multiple sheets.

PR #235, opened 2025-06-13
Crawling sitemap issues

ofermend:
fixed issues with blocked sitemap
Upgrade ray version
added custom user_agent that can be applied for crawling


adeelehsan (approved):
Looks good.

PR #242, opened 2025-06-30
Upgrade Docling

ofermend:
To prevent some strange warning  that was fixed in subsequent version of docling
also fixed a bug in csv crawler (CC @jcustenborder to make sure it this (minor) bugfix propagates also to your branch)


ofermend:
Implement in https://github.com/vectara/vectara-ingest/pull/266

PR #266, opened 2025-07-16
Improve PMC, FMP and Edgar crawlers

ofermend:
Bug fixes
Added parallelization


core/document_builder.py
```diff
@@ -0,0 +1,183 @@
...
+    
+    def _normalize_value(self, value):
+        """Normalize value using the provided normalize function"""
+        if isinstance(value, str):
+            return self.normalize_text(value)
+        return value
+    
+    def _generate_doc_id(self, doc_id: str) -> str:
+        """Generate a valid document ID, truncating if necessary"""
+        if len(doc_id) < 128:
```
adeelehsan: may be add constant instead of number ?

core/file_processor.py
```diff
@@ -0,0 +1,215 @@
...
+            with tempfile.NamedTemporaryFile(suffix=".pdf", mode='wb', delete=False) as f:
+                pdf_writer.write(f)
+                f.flush()
+                
+                pdf_part_id = f"{metadata['file_name']}-{i}"
+                pdf_parts.append((f.name, pdf_part_metadata, pdf_part_id))
+        
+        return pdf_parts
+    
+    def create_document_parser(self):
```
adeelehsan, line 87: To reduce the repetition and code we can extract some parameters into the base dict e.g 
```base_params = {
        'cfg': self.cfg,
        'verbose': self.verbose,
        'model_config': self.model_config,
        'parse_tables': self.parse_tables,
        'enable_gmft': self.enable_gmft,
        'summarize_images': self.summarize_images,
    }```

File: core/file_processor.py [hunk 3]
```diff
@@ -0,0 +1,215 @@
...
+                "start_page": i,
+                "end_page": min(i + pages_per_pdf, total_pages)
+            })
+            
+            # Add pages to writer
+            for j in range(i, min(i + pages_per_pdf, total_pages)):
+                pdf_writer.add_page(pdf_reader.pages[j])
+            
+            # Create temporary file
+            with tempfile.NamedTemporaryFile(suffix=".pdf", mode='wb', delete=False) as f:
```
adeelehsan, line 78: Since delete is set to false. Can we add clean up function to delete the temp files?

ofermend: It's cleaned outside this function in indexer.py

PR #274, opened 2025-08-11
adding saml support

adeelehsan:


crawlers/website_crawler.py
```diff
@@ -69,7 +162,8 @@ def crawl(self) -> None:
             all_urls = []
             for homepage in base_urls:
                 if self.cfg.website_crawler.pages_source == "sitemap":
-                    urls = sitemap_to_urls(homepage)
+                    # Pass SAML session for authenticated sitemap access
+                    urls = sitemap_to_urls(homepage, session=self.saml_session)
```
ofermend: do we need to add session to sitemap_to_urls? I don't think it has that yet and I don't see an update of spider.py in this PR

crawlers/website_crawler.py
```diff
@@ -131,14 +225,16 @@ def crawl(self) -> None:
             logger.info(f"Using {ray_workers} ray workers")
             self.indexer.p = self.indexer.browser = None
             ray.init(num_cpus=ray_workers, log_to_driver=True, include_dashboard=False)
-            actors = [ray.remote(PageCrawlWorker).remote(self.indexer, self, num_per_second) for _ in range(ray_workers)]
+            # Pass SAML session to workers
+            actors = [ray.remote(PageCrawlWorker).remote(self.indexer, self, num_per_second, self.saml_session) for _ in range(ray_workers)]
```
ofermend: Did you test this with Ray? I want to make sure it works in this case given that it starts new processes and make sure the SAML session serializes properly as such.

ofermend (changes requested):
Made some comments

ofermend (commented):
LGTM. Didn't test locally

PR #279, opened 2025-09-02
Refactor doc parsers

ofermend:


ofermend:
This PR is included in https://github.com/vectara/vectara-ingest/pull/280 so closing as "won't merge"

PR #283, opened 2025-09-16
adding hubspot crm crawler

adeelehsan:


core/indexer.py
```diff
@@ -424,29 +414,40 @@ def index_document(self, document: Dict[str, Any], use_core_indexing: bool = Fal
...
             logger.info(f"Can't serialize document {document} (error {e}), skipping")
             return False
 
-        try:
-            response = self.session.post(api_endpoint, data=data, headers=post_headers)
-        except Exception as e:
-            logger.info(f"Exception {e} while indexing document {document['id']}")
-            return False
+        # Define upload function for potential retry
+        def upload_document():
```
ofermend: Why do we need this to be a sub-function like this? Is it called more than one time in line 440?

ofermend: Oh I see - it's an argument in handle_file_upload_response
why is this type of retry needed? were there errors in the API calls? 
This seems a bit too much complexity. Why wouldn't a simple upload work?

crawlers/hubspot_crawler.py
```diff
@@ -1,41 +1,166 @@
...
+            logger.info("Starting HubSpot Email-Only Crawler.")
+            self._crawl_emails_only()
+        else:
+            logger.info("Starting Deal-Focused HubSpot CRM Crawler with Ray support.")
+            self._crawl_full_crm()
+
+    def _crawl_emails_only(self) -> None:
+        """Lightweight email-only crawling mode.
+
+        This mode only fetches contacts and their email engagements,
```
ofermend: Do we really need a mode to "imitate" the old code? Wouldn't there just be a mode that does email only and it just uses the same code only filters for emails?

crawlers/hubspot_crawler.py
```diff
@@ -131,7 +1677,807 @@ def get_contact_engagements(self, contact_id: str) -> Tuple[List[Dict[str, Any]]
...
+
+    def _build_ticket_document(self, ticket: Dict, associations: Dict) -> Dict:
+        """Build a simplified ticket document."""
+        # Delegate to the main class method
+        import sys
+        import os
+        sys.path.insert(0, os.path.dirname(os.path.dirname(__file__)))
+        from crawlers.hubspot_crawler import HubspotCrawler
+
+        temp_crawler = object.__new__(HubspotCrawler)
```
ofermend: Why do you name this "temp_crawler"?
Maybe just use "ticket_crawler"? same below with engagement, etc

PR #323, opened 2026-03-02
Pause/resume crawl

ofermend:
Support for pause/resume of a job
Fixed to image indexing from HTML pages


adeelehsan:
Currently, when a crawl is paused (via crawl_control.py pause or SIGTERM), the crawler enters an infinite polling loop in
   wait_if_paused() — sleeping 2 seconds and re-checking the DB forever. The process stays alive and the Docker container  
  keeps running, waiting for a resume signal.
                                                                                                                           
  This doesn't work well for a Docker-based workflow. If the container is paused, the only way to stop it is docker kill,  
  which is an ungraceful shutdown with no cleanup.
                                                                                                                           
  Suggested behavior: Pause should save progress to the DB, exit the process cleanly (container stops), and on restart the 
  crawler should detect the existing progress via get_indexed_ids() and pick up where it left off. The building blocks for
  resume are already there — we just need wait_if_paused() to signal the crawler to break out of its loop and exit rather  
  than spin indefinitely.

adeelehsan:
Can we go with simpler approach ? I think with new Cli tool we are complicating the things.
  - The crawler tracks every indexed/failed/skipped doc in a SQLite DB written to the output directory as it goes.
  - If the crawl is interrupted for any reason (crash, kill, stopped, OOM — doesn't matter how or where it's running),
  progress is already persisted.                                                                                           
  - On next run: if reindex: false, read get_indexed_ids() from the DB, skip already-processed docs, and continue from
  where it left off. If reindex: true, reset the DB and start fresh.

ofermend:
@adeelehsan thanks for the comments - accepted and revised. Can u pls review after the changes and test?

adeelehsan:
@ofermend  should we keep the flag `enable_crawl_tracking`? Don't you think by default enabling the tracking is a good idea ?

ofermend:
@adeelehsan just didn't want to make anything break so this makes it backwards compatible. We can always make it default on in a later version.

PR #338, opened 2026-05-14
box: don't infer deletions from incomplete folder listings

aamirbutt:
## Summary
- Fixes a bug where a single failed Box folder listing (e.g. a Box Shield 403 on the configured folder) caused the incremental crawl to delete every previously-indexed document from the Vectara corpus.
- Root cause: `_get_all_box_files` caught all exceptions from `folder.get(...)` / `folder.get_items(...)` and returned an empty/partial map. `_delete_removed_from_vectara` then treated every `file_id` in `indexed.csv` as "removed from Box" and deleted it.
- `_get_all_box_files` now returns `(files_map, listing_complete)`. The flag flips to `False` on any caught listing exception (top-level or recursive subfolder). `_crawl_incremental` gates the delete pass on the flag and logs a warning when it skips deletion. New-file indexing is unaffected — files we positively saw in Box are still indexed.

## Real-world impact
A Broadcom HR Box corpus running this code dropped from ~1505 documents to 16 in a single cron run after a Box Shield Access Policy started returning 403 `forbidden_by_policy` on the folder listing. With this change the same scenario would surface as a logged warning and leave the corpus intact until Box access is restored.

## Test plan
- [x] Syntax/import check (`python -c "import ast; ast.parse(...)"`).
- [ ] Run incremental crawl against a folder the credential can list — confirm normal flow (`listing_complete=True`, deletions happen as before).
- [ ] Run incremental crawl against a folder the credential cannot list (simulate 403) — confirm `listing_complete=False`, no deletions occur, warning is logged.
- [ ] Mixed-folder crawl where one of several `folder_ids` 403s — confirm deletion pass is skipped (conservative behavior).

### DIFF
--- a/crawlers/box_crawler.py
+++ b/crawlers/box_crawler.py
@@ -992,10 +992,10 @@ class BoxCrawler(Crawler):
         """
         all_files: Dict[str, Dict[str, Any]] = {}
         file_ext_set = set(str(e).lower() for e in self.file_extensions) if self.file_extensions else set()
-        listing_complete = True
 
-        def _crawl(fid: str, path: str = ""):
-            nonlocal listing_complete
+        def _crawl(fid: str, path: str = "") -> bool:
+            """Return True if this folder and all reachable subfolders listed successfully."""
+            complete = True
             try:
                 folder = self.client.folder(fid).get(fields=["name", "id"])
                 current_path = f"{path}/{folder.name}" if path else folder.name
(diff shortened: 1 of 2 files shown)


crawlers/box_crawler.py
```diff
@@ -978,16 +978,24 @@ def _parse_hours_back(hours_back_str: str) -> timedelta:
...
+        was removed from Box" from the returned map when the flag is ``False``,
+        because a transient listing failure would otherwise look like a deletion
+        of every file under the unreachable folder.
         """
         all_files: Dict[str, Dict[str, Any]] = {}
         file_ext_set = set(str(e).lower() for e in self.file_extensions) if self.file_extensions else set()
+        listing_complete = True
 
         def _crawl(fid: str, path: str = ""):
+            nonlocal listing_complete
```
ofermend: This is okay technically, but I usually try to avoid nonlocal. Is there a better way, maybe return listing_complete from _crawl() and then aggregate the result in the loop?

ofermend (approved):
LGTM, except a minor suggestion to avoid using nonlocal

crawlers/box_crawler.py
aamirbutt: Done — `_crawl` now returns whether it (and all reachable subfolders) listed successfully, and `_get_all_box_files` aggregates that in the loop. `nonlocal` is gone (df4cda3).

Also added `tests/test_box_crawler.py` (13a018f) covering `_get_all_box_files` for complete vs incomplete listings (top-level and subfolder failures) and the `_crawl_incremental` guard that skips the deletion pass when the listing is incomplete.

PR #354, opened 2026-07-09
Bundle docling/EasyOCR models for on-prem/air-gapped image builds

ofermend:
## Summary

Adds support for building `vectara-ingest` images for on-prem / air-gapped deployments, and fixes a `do_formula_enrichment` crash that surfaced while validating it.

By default `docling` fetches its layout and TableFormer models from HuggingFace Hub on first use, EasyOCR fetches its models from GitHub releases, and docling fetches NLTK data from GitHub — all fail with no outbound network access. A single build-arg bundles everything into the image at build time so no runtime fetch is attempted.

## Changes

**One build-arg bakes all offline models: `DOWNLOAD_DOCLING_MODELS`**
- Pre-bakes every model `docling_config` can select: default layout (heron) + `heron_101`/`v2` variants, TableFormer, and the CodeFormula model used by `do_formula_enrichment`. Sets `HF_HUB_OFFLINE`/`TRANSFORMERS_OFFLINE` so no HF call is attempted at runtime.
- `core/doc_parser.py` points `PdfPipelineOptions.artifacts_path` at the baked models when present, and falls back to the normal lazy HF fetch on default images.
- `artifacts_path` is deliberately hardcoded rather than read from `DOCLING_ARTIFACTS_PATH` — that exact env var name is already bound by docling's own pydantic-settings config as a global fallback, and setting it would break default (non-baked) image builds.
- CI publishes a third `:onprem` image tag alongside the regular and `.full` variants, with build-provenance attestation.

**Fix OCR on the baked image**
- Once `artifacts_path` is set, docling forces EasyOCR to load from `{artifacts_path}/EasyOcr/` with downloads disabled. The previous `DOWNLOAD_EASYOCR_MODELS` arg baked into `~/.EasyOCR/`, where docling no longer looks, so scanned-PDF OCR failed with `FileNotFoundError: craft_mlt_25k.pth` (air-gapped or not).
- EasyOCR models are now baked into `/opt/docling-models/EasyOcr/` via EasyOCR's own downloader (avoids coupling to docling's internal module layout, which relocated `EasyOcrModel` between versions).
- Recognizers are per-script. Baked set covers the world's most-used languages: craft detector + gen2 (Latin — en/fr/de/es/it/pt/nl/…, Simplified Chinese, Japanese, Korean, Cyrillic) + gen1 (Devanagari/Hindi, Arabic/Urdu, Bengali). `easy_ocr_config.lang` selects the script at runtime; scripts outside this set fail offline.
- Consolidated: `DOWNLOAD_EASYOCR_MODELS` and the standalone `download-easyocr-models.py` are removed; the single `DOWNLOAD_DOCLING_MODELS` flag now bakes docling + EasyOCR + NLTK.

**Pre-bake NLTK data + stop unconditional runtime downloads**
- Bake `punkt_tab` and `averaged_perceptron_tagger_eng` in the offline build step (`HF_HUB_OFFLINE` does not gate NLTK's GitHub fetch).
- `core/doc_parser.py` now downloads NLTK data only if not already present. `nltk.download()` always fetches its index over the network, so calling it unconditionally made every air-gapped run log a resolution error even with the data baked in.

**Bump transformers to fix `do_formula_enrichment` TypeError**
- `docling==2.94.0`'s CodeFormula/VLM loader calls `from_pretrained(..., dtype=...)`, but `transformers<4.56` forwards the unrecognized `dtype=` into the model constructor, crashing with `Idefics3ForConditionalGeneration.__init__() got an unexpected keyword argument 'dtype'`. Reproduces identically with live HF models — unrelated to the bundling work.
- Constrain `transformers[torch]>=4.56.0,<5` (4.56.0 is the first fixed release, verified by bisection). Only `transformers` (4.55.0→4.57.6) and `tokenizers` (0.21.1→0.22.2) change.

## Size impact

`regular` (`INSTALL_EXTRA=false`) build: base 4.04GB + docling models ~1.6GB + EasyOCR bundle ~750MB (mostly the three ~190MB gen1 scripts) + NLTK ~15MB ≈ ~6.4GB on-prem image. (The previous all-language EasyOCR bundle made this ~8.5GB.)

## Testing

- `do_formula_enrichment` succeeds end-to-end with `--network none` against the on-prem image.
- Offline (`--network none`) OCR verified: Latin OCR extracts text from a scanned PDF; all 8 baked scripts load and run with no download attempt; no NLTK errors.
- Verified `--network none` conversions with the default pipeline, both alternate layout models, and the base (non-baked) image.
- Not proven by automated tests: non-Latin OCR accuracy offline (requires real scanned samples per script) — the baked models load and run, but quality is EasyOCR's, not this change's.


adeelehsan:
Tested the on-prem image end-to-end: built with both build-args (8.56GB, matches the stated size) and ran docling parses inside `--network none` containers. The docling bundling itself works — default heron + TableFormer, `heron_101`, `v2`, and `do_formula_enrichment` all parse successfully fully offline (which also confirms the transformers >=4.56 fix). Found one issue:

**OCR fails on the baked image.** With both `DOWNLOAD_DOCLING_MODELS=true` and `DOWNLOAD_EASYOCR_MODELS=true`, parsing a scanned PDF with OCR enabled fails:

```
FileNotFoundError: Missing /opt/docling-models/EasyOcr/craft_mlt_25k.pth and downloads disabled
```

Root cause: once `artifacts_path` is set, docling looks for EasyOCR models under `{artifacts_path}/EasyOcr/` and disables downloads — but `DOWNLOAD_EASYOCR_MODELS` bakes them into `~/.EasyOCR/`, where docling no longer looks. So scanned-PDF ingestion fails on the on-prem image regardless of network access.

Fix appears to be one line in `docker/bin/download-docling-models.py`: pass `with_easyocr=True` to `download_models()`. The comment there is right that EasyOCR isn't HF-hosted, but docling's downloader fetches it from JAIDED into exactly the folder the runtime lookup expects. I verified this end-to-end: after downloading into `/opt/docling-models/EasyOcr/` inside a container and then severing its network, the scanned-PDF OCR parse passes (8 segments, 37s).

**One more thing worth a look — NLTK.** While auditing for runtime downloads (`docker diff` on a network-enabled run), I noticed ordinary docling parses fetch NLTK data (`punkt_tab`, `averaged_perceptron_tagger_eng`) from GitHub (`nltk/nltk_data`) at runtime — `HF_HUB_OFFLINE` doesn't gate that. Offline it degrades gracefully (identical segment counts in my runs), but every air-gapped run logs download errors and runs a degraded path, so it could be a potential issue for on-prem. Might be worth pre-baking that data too, e.g. `python3 -m nltk.downloader -d $HOME/nltk_data punkt_tab averaged_perceptron_tagger_eng` in the on-prem branch of the Dockerfile.

adeelehsan:
Looks good. One small change I think default id for sonnet would  `claude-sonnet-5`. Also we may need to delete temperature=0 from the call. can you please double check ?

adeelehsan (approved):

PR #360, opened 2026-07-19
fixed issue with conda build

ofermend:


ofermend:
Superseded by #365, which fixes the same failure by calling the `conda-build` console script instead of the `conda build` subcommand. That keeps conda-build in the activated env — one install per job, nothing added to `base`.

On the two Copilot suggestions here: collapsing the installs into a single `-n base` command would have broken this branch. `setup-miniconda` puts only `condabin/` on PATH, not `base/bin`, so `conda-verify` and `anaconda` installed into base are not reachable from the activated `test` env. The split was load-bearing.

PR #363, opened 2026-07-28
fix(run.sh): honor YAML boolean flags and abort on bad crawler_file

ofermend:
Three defects in `run.sh`, all silent — each produced a working-looking run that did the wrong thing. Found while building end-to-end coverage for the container build options.

## 1. Feature flags could not be read as YAML booleans

```bash
mask_pii=$(read_yaml_nested "data.get('vectara', {}).get('mask_pii', 'false').lower()")
```

`yaml.safe_load` turns `true` into a Python **bool**, which has no `.lower()`. The `AttributeError` is swallowed by the helper's `2>/dev/null || echo ""`, so the flag comes back empty and `needs_extra_features` sees nothing. Only a quoted `"true"` ever worked.

Consequence: `mask_pii: true` built the **base** image with `INSTALL_EXTRA=false`. `presidio-analyzer` lives in `requirements-extra.txt`, so `PRESIDIO_AVAILABLE` was false and `mask_pii()` returned the text unchanged after one log warning — **the PII got indexed**.

Reproduced end to end before the fix:

```
Building base image
docker buildx build --load --build-arg INSTALL_EXTRA="false" ... --tag="vectara-ingest:latest"
core.utils - WARNING - Presidio is not installed. PII masking is disabled.
```

with the SSN, credit card, and email all landing in the corpus. Fixed with `str(...)` before `.lower()`.

## 2. A missing or misnamed `crawler_file` did not stop the run

```bash
CUSTOM_CRAWLER_PATH=$(get_custom_crawler_path)
```

`get_custom_crawler_path` runs in a command substitution, so its `exit 9` / `exit 10` only ended the subshell. `run.sh` printed the error and then **launched the container anyway using the built-in crawler**, exiting 0 — a typo in `crawler_file` gave a silently wrong ingest. Codes 9 and 10 were unreachable.

Fixed with `|| exit $?` so the status propagates.

## 3. `summarize_tables` drove the image choice but is read by nothing

No Python reads `doc_processing.summarize_tables` — `parse_tables` is the real switch (`core/file_processor.py`, `core/indexer.py`, `core/doc_parser.py`). Table summarization also needs nothing from `requirements-extra`. Dropped from `needs_extra_features`, and corrected in `config/box-example.yaml` and the Box/Vertex docs, where three Vertex examples additionally had it — and `summarize_images` — nested under `folder_crawler` instead of `doc_processing`.

## Tests

Adds `tests/test_run_sh.py`: drives the real script with a stubbed `docker` on `PATH`, so it needs no daemon and never builds an image or starts a container (8 checks, ~4s).

- **4 of the 8 fail against the previous `run.sh`**, all 8 pass after.
- Full suite: 637 passed, 1 pre-existing failure (`test_pmc_crawler::test_pdf_url_uses_current_pmc_domain`, fails identically on clean `main`).
- `ruff` clean, `bash -n run.sh` clean.

Also verified against the end-to-end regression suite: `mask_pii: true` now selects `vectara-ingest-full` and masks SSN/card/email while leaving non-PII intact; exit codes 9 and 10 now fire without launching a container; and the image/table cases that switch to the full image for the first time as a result (`image-png`, `pdf-local`) still pass, as do the deterministic snapshot cases (`csv-got`, `folder-pdf`, `bulkupload`) with no drift.

### DIFF
--- a/config/box-example.yaml
+++ b/config/box-example.yaml
@@ -19,9 +19,8 @@ doc_processing:
   process_locally: true
 
   # Table handling
-  parse_tables: true           # Extract tables from PDFs
+  parse_tables: true           # Extract and summarize tables (summarization needs model_config)
   enable_gmft: false           # Disable GMFT to avoid OpenAI costs
-  summarize_tables: false      # Set to true to enable AI summarization (requires model config)
 
   # Image handling
   summarize_images: false      # Set to true to enable AI summarization (requires model config)
--- a/docs/box/BOX_INGESTION_GUIDE.md
+++ b/docs/box/BOX_INGESTION_GUIDE.md
@@ -43,8 +43,7 @@ vectara:
   output_dir: /tmp/box_inspection     # Where failed files go
 
 doc_processing:
-  parse_tables: true                  # Extract tables from PDFs
-  summarize_tables: false             # Skip AI table summarization
+  parse_tables: true                  # Extract and summarize tables from PDFs
   summarize_images: false             # Skip AI image summarization
 
 box_crawler:
--- a/docs/vertex-ai/README.md
+++ b/docs/vertex-ai/README.md
@@ -29,7 +29,7 @@ Vertex AI integration enables AI-powered table and image summarization during do
 
 ```yaml
 doc_processing:
-  summarize_tables: true
+  parse_tables: true      # extracts AND summarizes tables
   summarize_images: true
 
 table_model_config:
--- a/docs/vertex-ai/VERTEX_AI_SETUP.md
+++ b/docs/vertex-ai/VERTEX_AI_SETUP.md
@@ -112,10 +112,12 @@ vectara:
 crawling:
   crawler_type: folder
 
+doc_processing:
+  parse_tables: true      # extracts AND summarizes tables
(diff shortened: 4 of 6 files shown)


ofermend:
**CI note:** the one failing check is not from this PR.

`test_pmc_crawler.TestPmcPaperIndexing::test_pdf_url_uses_current_pmc_domain` fails identically on clean `main` (verified locally by stashing this branch's changes). It's stale from #361: `crawlers/pmc_crawler.py:115` deliberately dropped the `/pdf/` suffix —

```python
# PMC serves a JS interstitial at /pdf/ that yields no content for
article_url = f"https://pmc.ncbi.nlm.nih.gov/articles/PMC{pmc_id}/"
```

— but `tests/test_pmc_crawler.py:51` still asserts the old `.../PMC123/pdf/`. The test needs updating to match the intended behavior; that's a separate one-line change and I've left it out of this PR to keep the diff focused.

Everything else is green: 558 tests ran, that one failure, and all 8 new `tests/test_run_sh.py` checks passed on Linux.

adeelehsan (approved):

PR #368, opened 2026-08-11
Upgrade six vulnerable dependencies and fix ray's bundled aiohttp copy

abhishekpradhan:
## What this fixes

Security scans of images built on vectara-ingest flag six Python dependencies with known vulnerabilities. This upgrades all of them past their fixed versions:

| package | from | to |
|---|---|---|
| aiohttp | 3.14.2 | 3.14.3 |
| cryptography | 48.0.1 | 50.0.0 |
| pyopenssl | 26.2.0 | 26.4.0 |
| ray | 2.55.0 | 2.56.0 |
| pyasn1 | 0.6.3 | 0.6.4 |
| httplib2 | 0.22.0 | 0.32.0 |

It also fixes a subtle one: **ray bundles a private copy of aiohttp** (for its runtime-env agent), which security scanners flag separately from the main one. A new Dockerfile step upgrades that bundled copy to match, reading the version from requirements.txt the same way the existing torch step does — so it can never drift out of sync with the lock.

## Why pyopenssl moves too

Older pyopenssl declares it cannot work with cryptography 50, and Scrapy's TLS stack runs through pyopenssl — so the pair has to move together. 26.4.0 is the first release that accepts cryptography 50.

## Compatibility

Verified directly on a python:3.12 image carrying exactly this package set (built for vectara/platform#7177, which consumes this project's image):

- `pip check` is clean — the dependency set is as consistent as the current release.
- Everything imports and initializes: the TLS stack (Scrapy/Twisted + the new cryptography), ray (including the patched bundled-aiohttp path), transformers, and the Google Cloud clients.
- No other locked versions change — everything these six packages need is already satisfied by the current lock, which is why the lock diff is exactly six lines.

## Notes for reviewers

- `requirements.txt` was edited in line with the six `requirements.in` changes rather than fully recompiled, to keep the diff reviewable. A full recompile can follow separately.
- Downstream, vectara/platform#7177 pins these same versions as overrides until a release with this change ships — after that they become a redundant safety net.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,6 +51,17 @@ RUN if [ "$INSTALL_EXTRA" = "true" ]; then \
         python3 -m spacy download en_core_web_lg; \
     fi
     
+# ray vendors a private aiohttp for its runtime-env agent under
+# ray/_private/runtime_env/agent/thirdparty_files; scanners flag that copy
+# separately from the site-packages one. Same grep-from-lock pattern as the
+# torch pins above so it can never go stale on a lock bump.
+RUN AGENT_DIR=/usr/local/lib/python3.12/site-packages/ray/_private/runtime_env/agent/thirdparty_files \
+    && AIOHTTP_PIN="$(grep -E '^aiohttp==' requirements.txt)" \
+    && test -n "$AIOHTTP_PIN" \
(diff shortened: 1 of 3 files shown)


ofermend (approved):
LGTM

PR #369, opened 2026-08-11
fix: retry DELETE/PUT requests and make delete_doc fault-tolerant

adeelehsan:
## Problem

A website crawl with `remove_old_content: true` completes successfully, then the whole ingest run crashes at the very end with:

```
requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
```

The crash happens in the post-crawl cleanup (`_remove_old_content_if_needed` → `Indexer.delete_doc`), which issues sequential DELETEs to the Vectara API on a long-lived `requests.Session`. By that point, pooled keep-alive connections may have gone stale (idle-closed by the server/LB during the long crawl), and the first DELETE on a dead socket raises `RemoteDisconnected`.

Normally urllib3 would retry on a fresh connection, but `DEFAULT_RETRY_METHODS` in `core/utils.py` was `["HEAD", "GET", "OPTIONS", "POST"]` — DELETE excluded. For a non-retryable method, `Retry.increment` re-raises immediately (visible in the traceback: `retries.increment` → `raise reraise(...)`), so the configured `total=5` retries never apply. Nothing in `delete_doc` or the delete loop catches the exception, so one stale connection aborts the entire run after all crawl work is done.

## Fix

- **`core/utils.py`**: add `PUT` and `DELETE` (both idempotent, safe to retry) to `DEFAULT_RETRY_METHODS` so stale-connection/read errors are transparently retried with the existing backoff.
- **`core/indexer.py`**: wrap the delete call in `delete_doc` in `try/except requests.exceptions.RequestException` — a persistent failure on one doc now logs an error and returns `False` (the caller already skips-and-continues on `False`) instead of crashing the run.

## Testing

- Verified the mounted adapter now reports `allowed_methods: ['HEAD', 'GET', 'OPTIONS', 'POST', 'PUT', 'DELETE']`.
- `tests/test_utils.py` (64 passed) and `tests/test_indexer_utils.py` (21 passed).


ofermend (approved):
LGTM. Didn't test locally.

PR #373, opened 2026-09-15
deps: bump unstructured to 0.24.1 (CVE-2026-71428) and clear the remaining high alerts

adeelehsan:
## Why

`unstructured==0.18.26` carries **CVE-2026-71428** (critical, CVSS 9.3): `partition`, `partition_html` and `partition_md` fetch their `url=` argument without host validation, so a caller-controlled URL can reach loopback, internal services or cloud metadata. vectara-ingest only calls `partition_*(filename=...)`, so the path isn't reachable here, but every image built from this lock is flagged by container scanners (it is the last critical on the Broadcom VISTA crawler images, which build on `vectara/vectara-ingest`). 0.24.0 routes all `url=` fetches through a validating helper (`unstructured/safe_http.py`).

## What changes

`unstructured` 0.24 needs `unstructured-inference>=1.6.12`, which needs `torch>=2.10`, `onnxruntime>=1.25`, `opencv-python>=4.13.0.90`, `pypdfium2>=5`, `timm>=1.0.24`, and `unstructured` itself now needs `wrapt>=2.1.1`.

| | before | after |
|---|---|---|
| unstructured | 0.18.26 | 0.24.1 |
| unstructured-inference | 1.1.1 | 1.6.13 |
| torch / torchvision (CPU) | 2.7.1 / 0.22.1 | 2.13.0 / 0.28.0 (see the high-severity table below) |
| onnxruntime | 1.22.0 | 1.30.0 |
| opencv-python | 4.11.0.86 | 4.14.0.94 (capped `<5`) |
| pypdfium2 | 4.30.0 | 5.13.0 |
| wrapt / deprecated / opentelemetry-instrumentation | 1.17.2 / 1.2.18 / 0.61b0 | 2.4.1 / 1.3.1 / 0.65b0 |

Direct pins bumped because the new floors require them: `psutil` 6.1.0 → 7.2.2, `beautifulsoup4` 4.12.3 → 4.15.0, `matplotlib` 3.8.4 → 3.10.9, `markdown` 3.8.2 → 3.10.3, `python-docx` 1.1.2 → 1.2.0, and `pikepdf` 9.8.1 → `>=10.3.0,<11` in the extras.

Both locks recompiled with the Makefile command; the `nvidia-*`, `cuda-bindings` and `cuda-pathfinder` pins removed as the note in `requirements.in` asks. `effdet` and `pycocotools` drop out of the lock; nothing imports them. Code changes are limited to one chunking option (next paragraph); the APIs used (`partition_{pdf,html,docx,pptx,md}(filename=...)`, `chunk_by_title`, `chunk_elements`, `Table`/`Image`/`Title` elements, `metadata.text_as_html`/`page_number`) are unchanged.

**Chunking fix.** Since unstructured 0.22 the `by_title`/`basic` chunkers put every `Table` in its own chunk by default (`isolate_table=True`). `UnstructuredDocumentParser.parse()` skips standalone `Table` chunks in the chunked path (tables are indexed from the raw elements, and only when `parse_tables` is on), so after the bump **table text silently dropped out of the index for every chunked document with `parse_tables` off (the default)**. The parse comparison caught it (below). This PR passes `isolate_table=False` to `chunk_by_title`, `chunk_elements` and the partition-level chunking kwargs, restoring the pre-0.22 merge; output for the affected documents is again identical to 0.18.26. Two new tests assert table text survives `by_title` and `basic` chunking and fail without the change.

Other behaviour changes from unstructured's changelog: the hi_res default DPI is 350, `<pre>` in HTML becomes `CodeSnippet` elements, and library telemetry is off by default.

This PR also clears the remaining **high** Dependabot alerts on the lock, and takes torch past the version unstructured strictly needs so the same work isn't repeated next month:

| package | before | after | alert |
|---|---|---|---|
| transformers | 5.8.1 | 5.12.1 | GHSA-xrqw-3rrv-vx5w (high, `<5.10.0`; 5.10.0 was yanked, so the floor is 5.10.4) |
| tornado | 6.5.7 | 6.5.10 | GHSA-mpf4-983q-p7j4 (high, `<=6.5.7`) |
| Protego | 0.4.0 | 0.6.2 | GHSA-wjmf-p669-5m5p (high, `<=0.6.1`); arrives via scrapy, so it needs its own pin |
| pyarrow | 20.0.0 | 25.0.1 | GHSA-rgxp-2hwp-jwgg (high, `<23.0.1`); arrives via datasets/ray |
| torch / torchvision | 2.7.1 / 0.22.1 | 2.13.0 / 0.28.0 | GHSA-qfhq-4f3w-5fph (`<2.10.0`) **and** GHSA-rrmf-rvhw-rf47 (`<=2.12.1`) |

`nltk` (GHSA-8mgp-746c-j5xp, high) has **no fixed release**: the advisory covers every version through 3.10.3, which is the newest and what this lock pins.

**CUDA:** from torch 2.13 the PyPI Linux wheel declares a single `cuda-toolkit==13.0.3.0` dependency instead of the previous ~14 `nvidia-*-cu12` pins, so the old strip pattern missed it and it landed in the lock as a direct requirement. Both locks are now stripped of `nvidia-*` and `cuda-*`, and the note in `requirements.in` records both forms. Verified in the built image: `torch 2.13.0+cpu`, `torch.version.cuda is None`, and no `cuda-*`/`nvidia-*` distributions installed.

## Risks and mitigations

- **Table text in chunks** (see Chunking fix): without `isolate_table=False` this bump drops table rows from chunked text. Mitigated in this PR and covered by tests.
- **unstructured 0.24 downloads a spaCy model on first use**: its tokenizer (`unstructured/nlp/tokenize.py`) loads `en_core_web_sm` and, if it isn't installed, fetches the wheel from GitHub at runtime. With the new lock alone, parsing a DOCX or HTML file in a container with networking disabled fails with `RuntimeError: Failed to download spaCy model`. (The released 2.13.3 image isn't offline-capable either: 0.18.26 downloads NLTK `punkt_tab` at partition time and, offline, returns empty text for the same files. PDF `hi_res` still downloads its layout model from Hugging Face on first use, unchanged.) Mitigation in this PR: the Dockerfile installs `en_core_web_sm` 3.8.0 (the exact version unstructured hardcodes, pinned by sha256) in the builder stage for every variant and loads it at build time, so a missing model fails the build instead of a crawl. Net effect: DOCX/HTML/Markdown/PPTX parsing now works offline, which it did not in 2.13.3. Non-Docker installs (conda, pip from source) still download it on first use, the same way NLTK data is fetched today.
- **Heavier dependency set**: unstructured 0.24 adds spaCy (and its model stack) to the base requirements, so the regular image grows by spaCy plus the ~13 MB small model.
- **wrapt 2 / OpenTelemetry 0.65b0**: pulled in transitively; `pip check` is clean in both images and the full test suite passes.
- **transformers 5.8 → 5.12 and torch 2.7 → 2.13 move every model load** (gmft's table-transformer, docling layout models, whisper, easyocr). Mitigation: model-path checks run the same documents through gmft table extraction, the docling parser and a transformers load+forward on both stacks: 

| check (old stack → new stack) | 2.13.3 image (torch 2.7.1 / transformers 5.8.1) | this branch (torch 2.13.0 / transformers 5.12.1) |
|---|---|---|
| transformers load + forward (tiny DistilBERT) | ok, `[1, 26, 32]` | ok, `[1, 26, 32]` |
| gmft tables, procedure.pdf | 1 table, shape (3, 4) | 1 table, shape (3, 4) |
| gmft tables, attention.pdf | 5 tables: (4,5) (4,4) (10,5) (17,13) (12,3) | 5 tables: (3,5) (4,4) (10,5) (20,13) (12,3) |
| docling, procedure.pdf | 10 texts, 1 table, 462 chars | identical |
| docling, attention.pdf | 157 texts, 4 tables, 35,765 chars | identical |

Same table counts everywhere and identical docling output; on the attention paper two gmft tables come back with different row counts (4→3, 17→20) because table-transformer segments rows slightly differently under the newer transformers/torch. No failures on either stack.
- **Rollback**: revert the PR (lock bump, Dockerfile model, chunking fix, high-severity pins).

## Testing

**Docker image builds** from this branch (`linux/arm64`), both variants the release workflow publishes:
- `INSTALL_EXTRA=false` (the plain version tag): builds; `torch 2.13.0+cpu`, `torch.version.cuda is None`, no `nvidia-*`/`cuda-*` distributions installed, `pip check` clean, `import core.crawler, core.indexer; from unstructured.partition.pdf import partition_pdf` succeeds.
- `INSTALL_EXTRA=true` (`.full`): builds; with `--network none`, `en_core_web_sm` 3.8.0 and `en_core_web_lg` 3.8.0 both load, `unstructured.nlp.tokenize.sent_tokenize` works, `pip check` clean.

**Unit tests**, run inside the images (Linux, as CI):

| | `unittest discover` (CI's command) | `pytest tests/` |
|---|---|---|
| `main` in `vectara/vectara-ingest:2.13.3` | 571 OK (1 skipped) | 658 passed, 1 skipped |
| this branch in the new image | 573 OK (1 skipped) | 660 passed, 1 skipped |

The 2 extra tests are the new chunking tests; against the unfixed `doc_parser.py` both fail (`AssertionError: 'Network Observability 24.3.1' not found in 'Back Up the Data Aggregator\n\nSupported versions are listed below. Stop the data aggregator service before copying the files.'`). ruff (CI's pinned 0.15.14) reports no findings on the changed files before or after.

**Offline check**: with `docker run --network none`, `UnstructuredDocumentParser` with `by_title` chunking on DOCX, HTML, Markdown and PPTX all parse. Without the baked model the same call raises `RuntimeError: Failed to download spaCy model`; in the released 2.13.3 image it returns empty text (NLTK `punkt_tab` can't be downloaded).

**Parse comparison**, `UnstructuredDocumentParser` with `by_title` and `none` chunking on the same documents, old image vs new image (generated DOCX/PPTX/HTML/Markdown/PDF with a table, a numbered procedure and a shell command, plus two arXiv PDFs through hi_res):

| document (chunking) | text chars old → new | tables detected old → new | parse secs old → new |
|---|---|---|---|
| attention.pdf (by_title / none) | 39,765 → 41,115 / 36,259 → 36,497 | 4 → 4 | 151 → 114 / 61 → 96 |
| bert.pdf (by_title / none) | 63,764 → 64,335 / 60,778 → 60,921 | 9 → 9 | 85 → 138 / 84 → 132 |
| procedure.docx | 660 → 660 / 462 → 462 | 1 → 1 | 0.2 → 0.3 |
| procedure.html | 599 → 599 / 401 → 401 | 1 → 1 | 0.2 → 0.3 |
| procedure.md | 527 → 527 / 330 → 330 | 1 → 1 | 0.2 → 0.3 |
| procedure.pdf | 823 → 823 / 622 → 622 | 1 → 1 | 3.0 → 4.2 |
| review.pptx | 367 → 367 / 174 → 174 | 1 → 1 | 0.2 → 0.3 |

All 14 parses succeed on both stacks and the regression check (lost table/procedure/command text, lost tables, or >20% text shrink) passes. Before the chunking fix the same run failed: `by_title` text dropped by exactly the table rows (e.g. 660 → 470 for the DOCX) in all five generated documents. Remaining differences: `<pre>` in HTML/Markdown now yields `CodeSnippet` elements, and hi_res PDF parsing is roughly 1.5× slower on arm64 (DPI default 350).

https://claude.ai/code/session_01CTgTxTQQmTpG8XT1BX89bz


ofermend (approved):
LGTM

